PR #163, opened 2026-04-27
fix(meeting): skip empty orphan sessions in recovery discovery

moona3k:
## Summary

Fixes a noisy and misleading meeting-recovery path: launch recovery could show "interrupted recordings" for folders that only contain AAC init headers and no captured audio. Those sessions cannot be recovered, so discovery now removes obvious empty stubs before prompting the user.

The review pass also keeps completed meeting sessions safe when paths differ by macOS' `/var` and `/private/var` aliases. A stale lock should never cause a valid completed `meeting.m4a` folder to be purged.

## Recovery Flow

```text
orphan recording.lock
        |
        v
discoverPendingRecoveries()
        |
        +-- completed meeting row? -----> keep
        |                                 recover/discard cleans stale lock
        |
        +-- mic/system audio >= 4 KB? --> keep
        |                                 recovery can transcribe real audio
        |
        `-- empty stub ----------------> purge folder + lock
                                          no user-facing recovery dialog
```

## What Changed

- Filters orphaned sessions whose mic/system files are below the viable-audio floor.
- Auto-purges only those empty stub folders so they do not accumulate forever.
- Preserves mic-only/system-only recoveries when either channel has real audio.
- Preserves already-completed meeting sessions, including `/var` vs `/private/var` path aliases.
- Fails open on unexpected viability-check errors so discovery does not hide or purge user data.
- Adds focused recovery tests for empty stubs, mixed batches, single-channel recovery, completed-session preservation, and fail-open behavior.

## Test Plan

- [x] `swift test --filter MeetingRecordingRecoveryServiceTests`
- [x] `swift test`
- [x] GitHub CI `swift-test` on push and pull_request

### DIFF
--- a/Sources/MacParakeetCore/Services/MeetingRecordingRecoveryService.swift
+++ b/Sources/MacParakeetCore/Services/MeetingRecordingRecoveryService.swift
@@ -70,7 +70,7 @@ public final class MeetingRecordingRecoveryService: MeetingRecordingRecoveryServ
     ///
     /// Sessions whose audio is at or below this floor were almost always
     /// killed within a second of `MeetingRecordingService.startRecording`
-    /// writing the lock file but before any real audio frames were captured —
+    /// writing the lock file but before any real audio frames were captured -
     /// e.g., a force-quit, hard crash, or rapid dev-cycle restart. Surfacing
     /// these as "interrupted recordings" alarms the user over data that
     /// never existed and where `recover()` would error with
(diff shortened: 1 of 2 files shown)


moona3k:
Reviewer follow-up addressed in `39e103f`:

- Per-session viability checks now fail open: a DB/file-system error logs `meeting_recovery_viability_check_failed`, keeps the candidate visible, and does not purge user data.
- Empty-session purging now returns early when the folder is already gone, avoiding noisy purge failure logs.
- Added `testDiscoverKeepsCandidateWhenViabilityCheckFails` for the fail-open path.

PR #192, opened 2026-05-02
Fix double-tap dictation race in AudioRecorder lostRace check

moona3k:
## Summary

Fixes a race where quickly double-tapping Fn could make persistent dictation fail with:

`Recording failed: interrupted during subscribe`

This happens when the first provisional start is still starting up while the second, user-intended persistent start begins.

## Race Shape

```text
Before

Time ->

Start #1   set starting=true ---- await subscribe ---- stale, fails
Stop #1                         bump generation, clear starting
Start #2                                set starting=true ---- await subscribe
Start #1 defer                                      clear starting=false
Start #2 resume                                             sees stale flag, fails

Result: the replacement recording also fails.
```

```text
After

Time ->

Start #1   owns token A -------- await subscribe ---- stale, fails
Stop #1                         bump generation, clear starting
Start #2                                owns token B -------- await subscribe
Start #1 defer                                      token A is stale, no clear
Start #2 resume                                             token B still owns start, records

Result: the user-intended persistent recording stays alive.
```

## What Changed

- `AudioRecorder` now uses per-start ownership for the shared `starting` flag, so an older start cannot clear a newer start's in-flight state.
- `DictationService` now checks session ownership before mutating state after async start work, so a stale failed start cannot reset a replacement session to idle.
- Race tests now use explicit rendezvous/polling instead of fixed sleeps.

## Validation

- `swift test --filter AudioRecorderFormatChangeTests`
- `swift test --filter DictationServiceSessionTests`
- Full `swift test`: 2053 XCTest tests + 16 Swift Testing tests passed
- GitHub CI on head `7003fdb2`: both `swift-test` jobs passed

### DIFF
--- a/Sources/MacParakeetCore/Audio/AudioRecorder.swift
+++ b/Sources/MacParakeetCore/Audio/AudioRecorder.swift
@@ -69,6 +69,14 @@ public actor AudioRecorder {
     private var outputURL: URL?
     private var recording = false
     private var starting = false
+    /// Bumped on every `start()` entry that passes the entry guard. Each call
+    /// captures its value as `myStartCallGeneration`; the `defer` only clears
+    /// `starting` if no newer call has claimed the slot. Without this, an
+    /// aborted sibling's defer would clobber a legitimate replacement start
+    /// AND mislead `stop()` into taking its no-op path (which doesn't bump
+    /// `sessionGeneration`), letting the replacement claim a recording the
+    /// caller of `stop()` just asked to end.
+    private var startCallGeneration: Int = 0
 
     /// Minimum samples before sending to STT.
     /// FluidAudio requires at least 1 second of 16kHz audio (16,000 samples).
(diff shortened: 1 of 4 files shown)


moona3k:
## Fresh-eye review applied (commit 3adb4978)

Ran Codex + Gemini fresh-eye reviews in parallel. They disagreed:

- **Codex caught a real correctness hole** in the prior single-line fix. With `!self.starting` removed from `lostRace`, this sequence breaks:
  1. start #1 mid-subscribe (`starting=true`)
  2. STOP #1 Path A: bumps gen, `starting=false`
  3. start #2 enters: `starting=true`, awaits
  4. start #1 resumes, throws, defer clobbers `starting=false`
  5. STOP #2 fires: sees `starting=false, recording=false` → Path B (line 379-381) → throws WITHOUT bumping gen
  6. start #2 resumes: gen unchanged, lostRace=false → succeeds, claims a recording the caller of STOP #2 just asked to end
- **Gemini said the fix was sound** — missed that `stop()` reads `starting` to decide whether to bump generation, so a sibling defer clobber breaks `stop()`'s invariant in addition to the lostRace check.

Verified Codex was right by walking the trace. Window is microseconds and rare, but real.

## Revised fix

Restored `!self.starting` in lostRace. Replaced `defer { starting = false }` with a per-call guard keyed on a new `startCallGeneration` field — defer only clears `starting` if no newer call has claimed the slot. A sibling's defer leaves the active claim alone, which:
- Keeps lostRace's `!self.starting` check correct (defense-in-depth)
- Keeps `stop()`'s view of `starting` accurate so it always takes gen-bumping Path A when an in-flight start exists

## Test improvement (also from review)

Previous test relied on `Task.sleep(50ms)` for task #2 to enter the actor before release. If task #2 entered AFTER task #1 resolved, the test passed without exercising the bug — silent false-pass.

New sync uses `permissionProvider` as a deterministic gate: it's invoked from inside `start()` after `starting=true` is set, so signaling on the second call proves task #2 has entered. Verified by temporarily reverting the per-call defer — test fails with exact error `"start #2 should succeed... got: recordingFailed("interrupted during subscribe")"`. Restored the fix; all 2052 tests pass.

PR #220, opened 2026-05-04
Fix system-default microphone fallback

### LINKED ISSUE #218: Audio input test failed after v0.6.0 update (OBSBOT Meet 2 on macOS Tahoe 26.5)
After updating to v0.6.0 this morning, the microphone input test fails with:
"Audio engine failed to start: Microphone engine failed to start..."

• macOS: Tahoe 26.5
• Device: Mac mini M4 + OBSBOT Meet 2
• Works fine in System Settings > Sound > Input
• Confirmed regression: v0.5.6 works perfectly with the same mic/setup. Only v0.6.0 is broken.

 Permissions reset + coreaudiod killed, no change

Steps to reproduce: Open app → Audio Input settings → Test Input

Image

moona3k:
## Summary
Fixes a v0.6.0 microphone regression where some system-default USB/camera mics can fail the Audio Input test after MacParakeet explicitly sets the default CoreAudio device.

The fix keeps explicit routing for user-selected microphones, but restores AVAudioEngine's implicit System Default path before falling back to the built-in mic. This matches the v0.5.6 behavior that worked for the reported OBSBOT Meet 2 setup.

```text
Before
Selected mic -> explicit CoreAudio set -> start
System Default -> explicit CoreAudio set -> start/fail
Built-in -> explicit CoreAudio set -> start

After
Selected mic -> explicit CoreAudio set -> start
System Default -> AVAudioEngine implicit default -> start
Built-in -> explicit CoreAudio set -> start
```

## Tests
- swift test --filter MicrophoneCaptureTests
- swift test --filter Audio
- swift test (2223 tests, 1 skipped, 0 failures)

Fixes #218

### DIFF
--- a/Sources/MacParakeetCore/Audio/MicrophoneCapture.swift
+++ b/Sources/MacParakeetCore/Audio/MicrophoneCapture.swift
@@ -9,12 +9,52 @@ public struct MeetingInputDeviceAttempt: Equatable, Sendable {
         case builtIn
     }
 
+    public enum Routing: Equatable, Sendable {
+        case explicit(AudioDeviceID)
+        case implicitSystemDefault(resolvedDeviceID: AudioDeviceID?)
+    }
+
     public let source: Source
-    public let deviceID: AudioDeviceID
+    public let routing: Routing
 
     public init(source: Source, deviceID: AudioDeviceID) {
         self.source = source
-        self.deviceID = deviceID
+        self.routing = .explicit(deviceID)
+    }
+
+    private init(source: Source, routing: Routing) {
+        self.source = source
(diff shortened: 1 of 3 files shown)


moona3k:
Addressed the review feedback in 34383c84: the attempt model now separates source from routing, so public .systemDefault attempts with a concrete device ID remain explicit while MacParakeet's fallback builder uses a distinct implicit System Default route. I also added an injected input-device setter/start seam plus testPlatformSkipsInputDeviceSetterForImplicitSystemDefaultAttempt so the skip of setInputDevice for implicit default is directly covered.

PR #245, opened 2026-05-10
Add pause/resume to meeting recording (#235)

### LINKED ISSUE #235: [Feature Request] A Pause button for transcribing: same as Stop button, but stitches the recordings together into 1 long audio file
Hi all, thank you for creating such a useful tool. 

So the way I use live transcription a lot of times needs me to pause, read what I just said, and think of the next sentence.

The Pause function I'm speaking of is not different functionally from how Macparakeet works currently:

1. I press a button and speak. 
2. I press the button again and the transcription is pasted as text. 

The only difference is that at the end I want to have 1 long audio recording file instead of many small files. 

This way I can then easily re-transcribe that file using Whisper or another model in one go. 

So everything stays the same, only when you pause the new recording is appended to the previous recording until you press stop. 

Hope it makes sense! 
Thank you.

moona3k:
## Summary
Adds a Pause/Resume toggle to in-flight meeting recordings so users can take a thinking break without ending the session, then resume into the same audio file. The resulting `.m4a` has no silent gap, and persisted duration excludes paused time. Closes #235.

## For users
- Pause and resume an active meeting recording from three places: the floating pill (right-click menu), the meeting panel header, and the Transcribe-tab recording tile.
- The saved audio file is one continuous recording — pauses are invisible when you scrub it, so you can re-transcribe with Whisper or another model in one pass.
- Saved duration reflects only time actually recorded; pausing then stopping does not pad the number.
- Pause is visually quiet but honest: the rosette dims with a pause-bar overlay, the hover badge swaps the red dot for a pause glyph, and the status label flips to "Paused".

## How it works
Pause is a software toggle, not a capture teardown. The mic and ScreenCaptureKit streams stay subscribed; the service simply discards incoming buffers while paused. That keeps resume instant and avoids mic/system desync. The audio writer's PTS counter is preserved across the pause window, so the file is gap-free. The flow state machine is unchanged — pause is a sub-state of `recording`.

```
  [Recording] --pause--> [Paused] --resume--> [Recording]
       \                    |                     /
        \                   v                    /
         +-----------> [Stopped] <--------------+
```

## Code touchpoints
- `Sources/MacParakeetCore/Services/MeetingRecording/MeetingRecordingService.swift` — pause/resume actor methods, buffer-discard, PTS preservation
- `Sources/MacParakeet/App/MeetingRecordingFlowCoordinator.swift` — `togglePause()` with stale-toggle cancellation and 150ms reconciler
- `Sources/MacParakeetViewModels/MeetingRecordingPillViewModel.swift` and `MeetingRecordingPanelViewModel.swift` — `.paused` state plumbing
- `Sources/MacParakeet/Views/MeetingRecording/MeetingRecordingPillView.swift` and `MeetingRecordingPillController.swift` — dimmed rosette, pause-bar overlay, right-click menu item
- `spec/adr/014-meeting-recording.md` §11 — design rationale; REQ-MEET-011 in `spec/kernel/requirements.yaml`

Closes #235.


moona3k:
## 9-agent review pass — findings + fixes

Ran a fresh-eye review with 3 codex + 3 gemini + 3 claude agents on independent lenses (race conditions, audio pipeline, test coverage, UX consistency, concurrency safety, docs alignment, edge cases, backward compat, code style). Plus my own re-read.

### Critical bug found and fixed

**Live transcript silently died after first pause+resume cycle.** `pauseRecording` called `await captureOrchestrator.reset()` to clear partial joiner/chunker state. That reset zeroed `AudioChunker.totalSamplesProcessed`, so post-resume chunks emitted at `startMs ≈ 0`. `MeetingTranscriptAssembler.apply` deduplicates against `lastCommittedEndMs[source]` via `endMs > cutoff` — every post-resume word fell below the pre-pause cutoff and was filtered out. The audio file was fine (storage writer is independent of the chunker), and the post-stop final transcription is fine (re-runs the audio end-to-end), but the LIVE preview stopped advancing after the first pause.

Fix: stop calling `captureOrchestrator.reset()`. The cost is one boundary chunk that may concatenate pre-pause and post-resume samples in the live transcript only — at-most-5s artifact, not a correctness issue. Independently caught by Codex audio P0 and my own fresh-eye scan; the fix is a one-line removal plus a direct assembler regression test.

### Major issues found and fixed

- **Optimistic UI flip race with polling reconciler** (Codex races + my own). `togglePause` flipped `pillViewModel.state = .paused` before awaiting `meetingRecordingService.pauseRecording()`. The 150ms polling task could read `captureMode == .full` before the pause Task ran and revert pill state to `.recording`, producing a `.paused → .recording → .paused` flicker. Fix: store the toggle Task as `pauseToggleTask`, cancel stale ones on rapid clicks, flip the UI only after the service confirms.
- **Rapid-toggle reentrancy through `await captureOrchestrator.reset()`** (Claude edges). Same fix as above — removing the reset closes the suspension window where a resume Task could clear `paused` mid-pause and leave `accumulatedPausedDuration` undercounted.
- **Reduce Motion not honored on rosettes** (Gemini UX x2). Both `MerkabaPillIcon` (pill) and `SacredFlowerTile` (tile) animated unconditionally — accessibility regression. Now gate `isAnimating` on `!reduceMotion`.

### Minor UX/style fixes

- Hover scale 1.06 → 1.08 on `PauseResumeButton` (match `StopRecordingButton`).
- Tile rosette dim opacity 0.55 → 0.45 (match pill).
- Added pause-bars overlay to tile rosette (match pill).
- Right-click menu keeps `leaf` symbol across listening/paused (preserve flower metaphor; was swapping to `pause.circle`).
- Panel header status dot uses dimmed success-green when paused, not `textTertiary` (reads as "active but quiet" rather than "disabled").
- `guard let _ = currentSession` → `guard currentSession != nil`.
- Trimmed scattered `(issue #235)` comments and over-explained doc lines on new state vars.

### Tests added (43 total in MeetingRecordingServiceTests, was 40)

- `testCancelWhilePausedClearsAllPauseStateAndRecordingState` — cleanupState reset coverage.
- `testRapidPauseResumeStormSettlesInLastUserIntent` — idempotency under rapid toggle.
- `testLiveTranscriptDoesNotDieAfterPauseResume` — direct assembler regression guard against the captureOrchestrator-reset bug.

### Deferred (documented as known limitations in ADR-014 §11)

- **Sleep-during-pause auto-finalizes the meeting.** `sourceInterrupted` events aren't gated by `paused`, so a Mac going to sleep with a paused recording routes through the existing capture-failure path and ends the meeting cleanly. No data loss, but the user expecting to resume after wake gets a finalized meeting. Follow-up PR can either auto-pause-stop on `NSWorkspace.willSleepNotification` or attempt stream re-init on wake.
- **Telemetry semantic drift on `durationSeconds`.** `meetingRecordingCompleted` and `meetingRecordingCancelled` now emit active-recording time (excludes pause), not wallclock. Cohort analysis spanning the merge date will silently mix two definitions. The follow-up PR adding pause/resume telemetry events should add a `pausedSeconds` field on `meetingRecordingCompleted` so analyses can disambiguate; needs the website allowlist updated in lockstep.
- **Lock file lacks pause state.** Crash-during-pause leaves the lock state as `.recording`. Recovery's wallclock fallback over-counts pause time IF both source files are unreadable (edge case). Persisting `accumulatedPausedDuration` to the lock file would defend this.

### Validation

- `swift test` — 2264 tests, 1 skipped, 0 failures (added 3 new tests).
- `swift build` — clean.
- Diff: `eaf1fa8d..a1b6c7e6`, +194 / -88 across 11 files.

PR #270, opened 2026-05-11
Expose meeting audio file via Show in Finder and Save Audio As… (closes #262)

### LINKED ISSUE #262: Where to download/find the recording of the meeting? I like to have a little more control over when it's recorded so I can use it somewhere else too.
## Message

Where to download/find the recording of the meeting? I like to have a little more control over when it's recorded so I can use it somewhere else too.

## System Info

| Field | Value |
|-------|-------|
| App Version | 0.6.3 (20260505194542) |
| Git Commit | `058abb002c4f` |
| macOS | 26.2.0 |
| Chip | Apple M1 Pro |

---

moona3k:
## Summary
Meeting recordings always write the source `meeting.m4a` (plus separate `microphone.m4a` and `system.m4a`) to `~/Library/Application Support/MacParakeet/meeting-recordings/<UUID>/`, but until this PR the only GUI path to that audio was AVPlayer playback inside the detail view — the file itself was unreachable. Issue #262 surfaced this: a user on v0.6.3 asked *"Where to download/find the recording of the meeting? I like to have a little more control over when it's recorded so I can use it somewhere else too."* The "control over when it's recorded" half is mostly addressed by the pause/resume work in v0.6.4 (#235); this PR closes the **"use it somewhere else"** half.

## What's new
Two actions, gated on `sourceType == .meeting`, available in two surfaces:

| Surface | Where |
|---|---|
| Library Meetings list | Row right-click → **Show in Finder** / **Save Audio As…** |
| Meeting detail view | New **Audio** menu in the action bar (waveform icon), next to Export → **Show in Finder** / **Save Audio As…** |

Behavior:
- **Show in Finder** → `NSWorkspace.activateFileViewerSelecting` on the mixed track.
- **Save Audio As…** → `NSSavePanel` with `.mpeg4Audio` content type, default destination `~/Downloads`, smart filename (`<derivedTitle> - 2026-05-11.m4a` when the LLM-derived title is set, otherwise the recording's displayName). Copy runs on a `Task.detached` so multi-hundred-MB meetings don't stall the UI thread.
- Both items are **disabled with a tooltip** when the audio file isn't on disk (e.g. crash-recovered placeholder, deleted file).
- Detail-view success reuses the existing export-confirmation popover (now generalized from `format: String` to `title: String` so it can read "Saved Audio" without losing "Exported Markdown" wording for transcript exports).
- Library failures surface through a new "Save Failed" alert.

## Architecture
- `MacParakeetCore.MeetingAudioFile` — single seam between UI surfaces and the on-disk layout. Resolves the URL, checks availability, derives the filename. Testable without AppKit.
- `MacParakeet.MeetingAudioActions` — `@MainActor` AppKit wrapper. Thin.
- 13 new unit tests cover URL resolution, file-vs-directory disambiguation, derivedTitle preference, fileName fallback, whitespace handling, and path-separator sanitization.

## Out of scope (deliberate)
- File and YouTube transcriptions don't get these actions yet. File sources are already known to the user; YouTube downloads sit at `~/Library/Application Support/MacParakeet/youtube-downloads/` and could get the same treatment in a follow-up, but the immediate bug report was meeting-specific.
- Individual track export (mic-only / system-only). The three tracks already exist on disk; a follow-up could expose them via a "Save Track…" submenu.
- Drag-out from the Library row. Nice-to-have, more `NSItemProvider` edge cases.
- Telemetry. Adding `audioRevealedInFinder` / `audioSavedAs` events would require a coordinated change to `macparakeet-website`'s `ALLOWED_EVENTS` allowlist (see prior incident: the Worker rejects entire batches on unknown events). Worth doing if/when we want to measure adoption.

## Test plan
- [x] `swift test` — 2458 tests, 0 failures, 10 skipped (existing skips)
- [x] 13 new `MeetingAudioFileTests` cases pass
- [x] `swift build` clean on both Core and full GUI targets
- [ ] Manual: right-click a meeting in Library → Show in Finder reveals the `meeting.m4a` in its UUID folder
- [ ] Manual: right-click a meeting in Library → Save Audio As… defaults to Downloads with the derived-title filename and produces a playable .m4a copy
- [ ] Manual: in the meeting detail view, the new Audio menu appears for meetings only (not for file/YouTube transcripts) and both items work
- [ ] Manual: delete the underlying `meeting.m4a` and confirm both menu items become disabled with the help tooltip
- [ ] Manual: cancel the Save panel — no error alert, no spurious popover

Closes #262

## Summary by CodeRabbit

* **New Features**
  * Added "Show in Finder" action to access meeting audio files directly from the library and transcription detail views
  * Added "Save Audio As..." feature to export meeting recordings with custom location and naming
  * Automatic error alerts display when audio operations fail

[Review Change Stack](https://app.coderabbit.ai/change-stack/moona3k/macparakeet/pull/270)


moona3k:
Pushed fix-up commit `4bc3f1ca` addressing fresh-eye review findings (manual pass + independent Codex review).

## Issues caught and fixed

**🔴 P0 data-loss bug** — the original implementation did `removeItem(destination)` then `copyItem(source → destination)`. If a user picked the meeting's own folder as the save destination and confirmed overwriting `meeting.m4a`, the source would be deleted and the copy would then fail — losing the only recording. Fixed via new `MeetingAudioFile.safeCopy(from:to:)` that (a) no-ops on same-file save and (b) writes via a UUID-suffixed temp sibling + atomic `replaceItemAt`, so a mid-copy failure never corrupts the prior destination.

**🟡 Timezone-flaky test** — `MeetingAudioFileTests.makeDate` pinned dates to `America/Los_Angeles` while the production formatter uses system tz. Reproduced the failure with `TZ='Asia/Tokyo' swift test` → expected `2026-05-11`, got `2026-05-12`. Production behavior left unchanged (system-tz is the right UX — a meeting recorded at 11:30 PM Friday locally should show "Friday's date" in the filename, not roll over to UTC's Saturday). Test now uses `Calendar.current` to match.

**🟡 Library `.sourceUnavailable` silent no-op** — both reviewers flagged this independently. The Library path discarded the outcome enum, so if the audio file disappeared between menu render and click, nothing happened. Now surfaces the existing "Save Failed" alert.

**🟡 Library save success had zero feedback** — detail view played a sound + showed a popover; Library was silent. Now plays the same `transcriptionComplete` sound (popover anchoring on a scrolling row is fiddly; sound is the minimal-friction signal).

**🟢 Filename hardening** — extended `sanitize` to strip `CharacterSet.controlCharacters` (tabs, NULs, BEL, etc.) on top of `/:\"`. Added a 100-grapheme-cluster cap via `String.prefix(_:)` so an unusually long LLM-derived title can't produce a path the Save panel accepts but the filesystem rejects. Bidi formatters deliberately left intact so Arabic/Hebrew titles render correctly.

**🟢 Documented `isAvailable` semantics** — status-agnostic by design. A `.processing` or `.error` meeting whose audio is on disk should remain reachable from the GUI (ADR-019: fragmented MP4 is playable up to the last fragment). Added an inline comment so future readers don't add a status gate without an explicit product reason.

## Quality gates
- `swift test`: 2465 tests, 0 failures, 10 skips (added 7 new tests: control-char sanitization, stem-length cap, and 5 `safeCopy` cases)
- `swift test` with `TZ='Asia/Tokyo'` and `TZ='Australia/Sydney'`: clean
- `swift build`: clean

## Items I considered but did not change

- **`Menu` + `.parakeetAction(.secondary)` styling** — the two reviewers contradicted each other (Explore said fine, Codex flagged "may produce double-border on macOS 14 vs 15"). I trust Explore's read of the actual style implementation but flagging this for visual verification during manual testing.
- **`isAvailable` not gating on `.completed` status** — deliberate per ADR-019. Documented inline.
- **Bidi formatter sanitization** — declined; would break legitimate non-Latin titles.

PR #274, opened 2026-05-12
Remove "Best available" YouTube audio quality (CLI 3.0.0)

moona3k:
## Summary

- Removes the **Best available** YouTube audio quality option from the GUI, the CLI, and `YouTubeAudioQuality` in `MacParakeetCore`.
- Bumps `macparakeet-cli` to **3.0.0** (breaking removal of an enum value on `transcribe --youtube-audio-quality` and `config set youtube-audio-quality`).
- Folds the previously-Unreleased CLI work (`--engine app-default`, `--speaker-detection app-default|on|off`, expanded `config`) into the 3.0.0 release entry.
- 217+ focused tests pass locally (`swift test`).

## The bug behind the cut

A user opened a YouTube transcription and the audio scrubber's play button did nothing. Investigation showed the saved audio file was `.webm` (Opus codec). **AVPlayer on macOS has no native WebM demuxer and no native Opus decoder** — so the in-app audio scrubber silently failed while the video panel still worked (because the video panel re-extracts a fresh playable HLS/MP4 stream URL via yt-dlp on demand).

That mismatch is exactly the "Best available" path: `yt-dlp -f bestaudio/best` almost always picks YouTube's Opus-in-WebM stream because it's the highest bitrate the API exposes.

## Why removal beats the alternatives

We worked through three other options before landing here. None of them earn their keep:

| Option | Problem |
|---|---|
| **Transcode Opus → AAC at download time** (~3–5s for 1h video via bundled ffmpeg) | Adds a second generation of lossy encoding. After the transcode, "Best available" is *worse* than `m4a` (one generation from YouTube's master vs two), so the option's name becomes dishonest. |
| **Stream-URL fallback for unplayable local files** | Violates local-first. Re-introduces yt-dlp rate-limit exposure on every detail-view open — the exact reason `MediaPlayerViewModel.prepare()` vs `load()` were split (see the comment at `MediaPlayerViewModel.swift:59`). Breaks offline playback. Fails entirely for taken-down / geo-blocked videos. |
| **Sharper warning copy** | We actually shipped this to main as `f6edccc7`. It's a band-aid: users skim Settings detail text and still hit the dead play button. This PR supersedes that commit. |

**STT quality is unaffected** by this change — Parakeet sees 16 kHz mono WAV after the FluidAudio AudioConverter step regardless of the source format. The "Best available" mode never improved transcription accuracy; it only changed the *saved file* format.

## Why a MAJOR bump

`macparakeet-cli` 2.1.0 introduced `--youtube-audio-quality best-available` on **2026-05-10** (one day before this PR). Per `Sources/CLI/CHANGELOG.md`'s compatibility policy, removing an accepted flag value is a MAJOR change. The CLI surface is young enough that a clean 3.0.0 cut is more honest than a deprecation window for a value that's been live for ~24 hours and was broken from day one.

## Migration / backward compatibility

- **Existing UserDefaults values of `best_available`** silently coerce to `.m4a` on read. `YouTubeAudioQuality.current()` already returned `.m4a` for unknown raw values; the removal makes `best_available` "unknown." No migration code needed.
- **CLI scripts passing `--youtube-audio-quality best-available`** will now exit with code 2 (validation error). The error message points at `m4a` as the only supported value.
- **`config set youtube-audio-quality best-available`** throws `ValidationError("...Use m4a.")`.
- **Previously-downloaded `.webm` audio files** are left on disk. Their transcripts work; their video panels work (yt-dlp re-extracts a stream); their audio scrubber stays mute. This is a known limitation, not a regression — they were already broken. Cleanup affordances are out of scope for this PR.

## Caveat (not a regression)

With `m4a` quality the yt-dlp format selector remains `bestaudio[ext=m4a]/bestaudio/best`, so a rare YouTube video that exposes *no* m4a stream at all can still produce webm via the fallback. The audio scrubber is broken for that rare case, same as before. Eliminating that edge case would require either a hard `m4a`-only selector (which fails entirely on m4a-less videos) or the transcode path (rejected above). We'd address it case-by-case if it shows up in the wild.

## Test plan

- [x] `swift build` (clean)
- [x] `swift test` (full suite, exit 0)
- [x] Focused tests: `ConfigCommandTests | TranscribeCommandTests | YouTubeDownloaderTests | SettingsViewModelTests` (217 tests, 0 failures)
- [x] Manual: open `Settings → Transcription` and confirm "YouTube audio quality" row is gone (previous row was the YouTube transcription hotkey, next row is Speaker detection)
- [ ] Manual (post-merge sanity): transcribe a YouTube video, confirm the saved file is `.m4a` and the in-app audio scrubber plays

## ADRs / spec changes

No ADRs touched. `spec/05-audio-pipeline.md` updated to describe only the m4a-first selector and reference `Sources/CLI/CHANGELOG.md` for the removal rationale. `docs/cli-testing.md` updated to drop `best-available` from the CLI tree + example transcribe command.

## Commit

See `9bfef1ab` for the full rationale in the rich commit message (`What Changed` / `Root Intent` / `Prompt That Would Produce This Diff` / `Files Changed`).

## Release Notes

* **Breaking Changes**
  * Removed `--youtube-audio-quality best-available` option; only `m4a` is now supported
  * Removed YouTube audio quality picker from Settings UI

* **New Features**
  * Added `--engine app-default` flag to use shared app defaults for speech engine
  * Added `--speaker-detection` option with `app-default|on|off` modes (with alias `--no-diarize`)
  * Expanded CLI config commands to manage transcription defaults (engine, language, speaker detection, save audio)

[Review Change Stack](https://app.coderabbit.ai/change-stack/moona3k/macparakeet/pull/274)

### DIFF
--- a/Sources/CLI/CHANGELOG.md
+++ b/Sources/CLI/CHANGELOG.md
@@ -80,6 +80,38 @@ by checking exit code first: `2` = misuse, `1` = runtime, `0` = success.
 
 ## [Unreleased]
 
+## [3.0.0] -- 2026-05-11
+
+### Removed (breaking)
+
+- `transcribe --youtube-audio-quality best-available` is no longer accepted.
+  ArgumentParser rejects the value at parse time with a usage error (exit
+  code 2). The flag still accepts `app-default` and `m4a`; `m4a` is the only
+  saved-file format going forward.
+- `config set youtube-audio-quality best-available` is rejected with a
+  `ValidationError` ("Use m4a."). `config get youtube-audio-quality` returns
+  `m4a` for any legacy stored value, including the now-unknown
+  `best_available` raw string.
+- `YouTubeAudioQuality.bestAvailable` enum case removed from
+  `MacParakeetCore`. Existing UserDefaults values of `best_available`
+  silently coerce to `.m4a` on read; no migration code is needed.
+
+  **Rationale.** AVPlayer on macOS cannot decode WebM or Opus, the formats
+  yt-dlp produces under "best available." The saved audio file silently
+  failed to play through the in-app audio scrubber while the video panel
+  worked (because the video panel re-extracts a fresh playable stream URL).
+  The option's only nominal benefit — a marginally higher upstream Opus
+  bitrate — disappears after the 16 kHz mono WAV downsample used for STT,
+  so transcription quality is unaffected by this removal. We considered
+  three alternatives before the cut: (1) transcoding Opus → AAC at download
+  time, which adds a second generation of lossy encoding and makes "best
+  available" worse than `m4a`; (2) falling back to a streamed YouTube URL
+  for audio playback, which violates the project's local-first stance,
+  reintroduces yt-dlp rate-limit exposure on every detail-view open, and
+  breaks offline playback; (3) keeping the option with a sharper warning
+  string, which leaves the trap intact for users who skim Settings copy.
+  Removal is the only path that respects "Simplicity is the product."
+
 ### Added
 
 - `transcribe --engine app-default` resolves the speech engine and Whisper
--- a/Sources/CLI/Commands/ConfigCommand.swift
+++ b/Sources/CLI/Commands/ConfigCommand.swift
@@ -27,7 +27,7 @@ struct ConfigCommand: ParsableCommand {
           whisper-language          auto|<Whisper language code>    default: auto
           speaker-detection         on|off                          default: off
           save-transcription-audio  on|off                          default: on
-          youtube-audio-quality     m4a|best-available              default: m4a
+          youtube-audio-quality     m4a                             default: m4a
 
         Full event catalog:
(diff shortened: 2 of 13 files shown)


moona3k:
Superseding this PR — the removal premise is wrong.

I missed issue #237, which has the empirical measurement that drove the `best_available` option in the first place. On the Steve Jobs Stanford speech, `best_available` (Opus) gave **7.30% WER** vs `m4a` (AAC) at **8.13%** — a ~10% relative WER reduction and 33 word-level edits between the two transcripts.

So the option's headline benefit (better transcripts) is real. The bug is that the saved file doesn't play through AVPlayer.

New plan: keep `best_available`, fix playback by transcoding the saved webm → m4a in a background task right after STT completes. Single resulting file, no double-storage, lazy migration for existing webm-backed transcriptions.

Follow-up PR incoming.

PR #304, opened 2026-05-18
Subtitle quality: clause-start breaks, CPS guard, 5-cue LLM window, AI refinement UI

justingonz96-creator:
## Summary

- **Clause-start break heuristic** — `buildSubtitleCues` now proactively breaks before subordinating conjunctions and relative pronouns (`because`, `although`, `since`, `while`, `whereas`, `unless`, `until`, `though`, `who`, `which`, `where`, `when`) even without sentence-ending punctuation. The `findBestBoundaryBackward` scorer also awards a +175 bonus when a candidate split lands right before one of these words, giving them net positive weight even after the existing `badStarters` penalty.
- **CPS guard** — new `enforceReadingSpeed()` post-processing step splits cues that exceed `maxCPS` (default 17.0 chars/sec — the Netflix/BBC professional standard). If no clean split exists, the cue is left untouched. New `maxCPS: Double = 17.0` field on `SubtitleExportConfig`.
- **5-cue LLM refinement window** — `SubtitleLLMRefiner` expanded from a 3-cue sliding window (prev + current + next) to 5 cues (2 prev + current + 2 next), giving the LLM broader context for boundary decisions.
- **AI Refinement UI** — "Use AI Refinement" toggle added to the subtitle config section in the export popover, wired to async SRT/VTT export paths via new `exportToSRT(…llmService:)` / `exportToVTT(…llmService:)` overloads and `exportTranscriptToDownloadsAsync`.

## Why

Subtitle cues were breaking in linguistically awkward places — mid-clause, before dependent clauses, ignoring natural subordination. The clause-start heuristic adds zero-cost breaks at real linguistic boundaries. The CPS guard enforces a reading-speed floor so fast-spoken segments stay readable. The wider LLM window enables the optional AI pass to catch boundary issues that span more than one adjacent cue.

## Reviewer notes

- The two pre-existing test failures (`testBuildSubtitleCuesBreaksOnLongGap` + a related fatal crash) were present before this branch and are tracked separately — they are not regressions from this PR.
- `maxCPS` is config-only for now (no UI slider); the default of 17.0 matches professional broadcast standards and can be set to 0 to disable.
- `useLLMRefinement` defaults to `false` — opt-in, no behavior change for existing exports.

## Test plan

- [ ] Export an SRT from a long transcription — verify breaks land before `because`/`although`/`since` etc. rather than mid-clause
- [ ] Verify fast-spoken segments (dense word packing, short duration) get split by the CPS guard
- [ ] Enable "Use AI Refinement" in export options and confirm the async path fires
- [ ] Run `swift test` — all tests pass except the two pre-existing failures noted above

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/argmaxinc/argmax-oss-swift",
       "state" : {
-        "revision" : "e2adabbe7d98dc4d0ab9a5b75424ecc42a9cdbef",
-        "version" : "0.18.0"
+        "revision" : "25c62997041c134b03ca82731ce2f6fd2cae1eb9",
+        "version" : "1.0.0"
       }
     },
     {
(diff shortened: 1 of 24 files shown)


justingonz96-creator:
Closing in favour of a clean rebase onto current main — the upstream diverged significantly. Reopening as a focused single-commit PR.

PR #318, opened 2026-05-22
Calendar auto-start reliability hardening (pre-enable)

moona3k:
Closes the correctness gaps two independent audit passes found in the calendar-driven meeting auto-start feature (ADR-017), so it's safe to enable. **The `AppFeatures.calendarEnabled` flag is intentionally left OFF in this PR** — this makes the feature *correct*; flipping it is a separate decision after the hands-on E2E pass the flag comment calls for.

## Why
Marketing MacParakeet as a Granola alternative needs calendar auto-start ("you never press record"). The audits found no crash/data-loss/leak/telemetry blockers, but a cluster of correctness gaps concentrated in auto-stop and multi-meeting transitions — exactly the back-to-back-meeting power-user path. Round 2 also caught a reachable **privacy** defect round 1 missed.

## What's fixed (commit-by-commit)
1. **Idempotent auto-stop (privacy must-fix)** — auto-stop confirm was a blind `toggleRecording()`. If you manually stopped the auto-started recording during the 30s countdown, completion *started a brand-new* mic + system-audio recording. Now routes through `stopFromCalendar()` (never starts) + dismisses the stale toast on external stop + re-verifies ownership at fire time.
2. **Parser precision** — generic fallback no longer matches `call` in `recall`/`teamcall`/`?utm=fallcall` or `video` in `videogame.com`; adds Zoom `/w/` `/my/` + `zoomgov.com` + Whereby.
3. **RSVP gating** — don't auto-record `.declined`/`.pending` invites; reminders stay lenient; auto-stop is not RSVP-gated.
4. **Reschedule re-fire** — suppression sets key on `(id, startTime)` so a same-day move re-fires.
5. **Auto-stop reliability** — poll on `NSWorkspace.didWake`, re-inject the owned event when EventKit drops it past `endTime`, and a 5-min forgiveness tail so sleep / long meetings still auto-stop.
6. **Poll reentrancy guard** — `EKEventStoreChanged` bursts can no longer interleave polls into duplicate reminder notifications.
7. **Back-to-back retry** — a meeting rejected as `state_busy` retries once the blocking recording ends (within its auto-start window).
8. **Notification-permission UI** — Settings now surfaces "Notifications are off" so `.notify` can't be a silent no-op; also fixes a stale header comment.
9. **Zero-duration event filter.**

## Tests
Full suite green: **2800 XCTest (10 skipped, 0 failures)** + all Swift Testing suites. Added coverage: idempotent/ownership-recheck auto-stop, parser variants & false-positives, RSVP gating, reschedule re-fire, forgiveness window, owned-event merge, deterministic poll-reentrancy (new mock fetch gate), back-to-back retry, and a state-machine invariant that `.stopRequested` from idle never starts.

## Not in scope (deliberate)
- **Flipping `calendarEnabled`** — still OFF. Recommend: one more focused audit pass on this diff, then the hands-on E2E (CLI read-path against a real calendar + a real Zoom/Meet run incl. the full-screen-toast check), then flip with `.notify` as the safe default.
- Late-join after a missed auto-start window (Phase 3).
- Observed-but-unrelated: the reminder picker's "At start time" maps to `reminderMinutes = 0`, which disables reminders — pre-existing, worth a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added calendar notification authorization checking with quick access to notification settings.
  * Improved meeting link detection for Zoom variants (personal rooms, webinars, GOV) and Whereby.
  * Extended auto-stop window with forgiveness period past event end time.
  * Added RSVP status filtering for auto-start (skips declined/pending invites).

* **Bug Fixes**
  * Prevented accidental recording restart when stopping auto-started recordings.
  * Fixed reminders/auto-start to properly track rescheduled event occurrences.
  * Improved handling of auto-start/auto-stop after Mac sleep/wake.
  * Eliminated duplicate reminder notifications from concurrent polling.
  * Filtered invalid calendar events with invalid date ranges.

[Review Change Stack](https://app.coderabbit.ai/change-stack/moona3k/macparakeet/pull/318?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/MacParakeet/App/AppEnvironmentConfigurer.swift
+++ b/Sources/MacParakeet/App/AppEnvironmentConfigurer.swift
@@ -323,7 +323,10 @@ final class AppEnvironmentConfigurer {
                     meetingCoordinator?.startFromCalendar(title: title)
                 },
                 onAutoStopConfirmed: { [weak meetingCoordinator] in
-                    meetingCoordinator?.toggleRecording()
+                    // Idempotent stop — never a toggle. A toggle would *start*
+                    // a new recording if the user already stopped the
+                    // auto-started one during the 30s auto-stop countdown.
+                    meetingCoordinator?.stopFromCalendar()
                 }
             )
             // The recording flow tells the calendar coordinator when an
--- a/Sources/MacParakeet/App/AppEnvironmentConfigurer.swift
+++ b/Sources/MacParakeet/App/AppEnvironmentConfigurer.swift
@@ -322,11 +322,11 @@ final class AppEnvironmentConfigurer {
                 onAutoStartConfirmed: { [weak meetingCoordinator] title in
                     meetingCoordinator?.startFromCalendar(title: title)
                 },
-                onAutoStopConfirmed: { [weak meetingCoordinator] in
+                onAutoStopConfirmed: { [weak meetingCoordinator] recordingGeneration in
                     // Idempotent stop — never a toggle. A toggle would *start*
                     // a new recording if the user already stopped the
                     // auto-started one during the 30s auto-stop countdown.
-                    meetingCoordinator?.stopFromCalendar()
+                    meetingCoordinator?.stopFromCalendar(recordingGeneration: recordingGeneration)
                 }
             )
             // The recording flow tells the calendar coordinator when an
--- a/Sources/MacParakeet/App/MeetingAutoStartCoordinator.swift
+++ b/Sources/MacParakeet/App/MeetingAutoStartCoordinator.swift
@@ -40,8 +40,8 @@ final class MeetingAutoStartCoordinator {
     /// an auto-start recording. The event title is forwarded so the
     /// recording flow can pre-name the saved transcription with the
     /// calendar event name instead of the date-based default.
-    private let onAutoStartConfirmed: @MainActor (_ title: String) -> Void
-    private let onAutoStopConfirmed: @MainActor () -> Void
+    private let onAutoStartConfirmed: @MainActor (_ title: String) -> Int?
+    private let onAutoStopConfirmed: @MainActor (_ recordingGeneration: Int) -> Void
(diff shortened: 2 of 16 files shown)


moona3k:
## Follow-up: cross-model review of the diff (+4 fixes)

Ran an independent Codex + Gemini review over the cumulative diff (the convergence pass), plus a third forwarded review. Verified each finding against the code rather than relaying — several were non-issues. Landed 4 real fixes:

- **`8b4a655c`** — poll hardening: re-read mode/permission *after* the fetch await (a mode→off during the fetch was processed once); coalesce dropped polls so a settings change mid-fetch isn't lost; keep 5s cadence while an auto-stop is owed (missed-window catch-up). Strengthened the reentrancy test to assert the coalesced re-run.
- **`4c8211ab`** — generic meet-link match must be in host/path, not query string (`?type=conference` on a Jira/CRM link no longer counts).
- **`6c17db4a`** — Settings "Grant Calendar Access" now defaults to `.notify` (matched onboarding; granting did nothing before).

### Verified non-issues (no change)
- "Dismissed auto-stop toast re-fires every poll" — REFUTED: `dismissedEventIds` is keyed by `dedupeKey` and filters the candidate before the auto-stop block, so it can't re-fire (even across reschedule).
- Onboarding footer Continue not calling `skipCalendarStep()` — matches the existing `.meetingRecording` step pattern; result is mode `.off` either way.

### E2E status
The CLI `calendar upcoming` read-path can't be validated headlessly — it deliberately defers Calendar permission to the GUI (returns `.notDetermined`), and the CLI binary is a separate TCC identity. Real read-path E2E (EventKit conversion, real Zoom/Meet, full-screen toast) still needs a GUI run with the flag flipped locally. **Flag remains OFF in this PR.**

Full suite green: **2802 XCTest, 10 skipped, 0 failures.**

PR #332, opened 2026-05-23
Reuse activation timestamp formatter

moona3k:
## Summary
- addresses Gemini's PR #330 review by reusing the first-dictation activation timestamp parser
- wraps the shared ISO8601DateFormatter behind a lock because it is mutable/non-Sendable and the callback is @Sendable

## Validation
- git diff --check
- swift test


moona3k:
Closing this as intentionally not worth carrying forward. The original Gemini suggestion is technically reasonable in isolation, but this callback runs only on the first successful dictation per install, so avoiding one ISO8601DateFormatter allocation is not worth adding a locked @unchecked Sendable wrapper.

PR #337, opened 2026-05-23
Expose meeting result availability in CLI

moona3k:
## Summary
- Add `hasPromptResults` and `promptResultCount` to meeting list/show/export JSON.
- Show prompt-result counts in human meeting list/show output.
- Add repository count coverage plus CLI spec and changelog updates.

## Tests
- `swift test --filter MeetingsCommandTests`
- `swift test --filter PromptResultRepositoryTests`
- `swift test --filter SpecCommandTests`
- `git diff --check`
- `swift test`


moona3k:
Follow-up pushed in 3e550d3 addressing the Gemini/Greptile review items: prompt-result counting is now part of the repository protocol, meeting list uses one grouped bulk count query instead of per-row reads, and Markdown export metadata includes the prompt-result count. Re-ran targeted tests plus the full Swift test suite locally.

PR #341, opened 2026-05-23
Fix dictation hotkey defaults and recorder persistence

moona3k:
## What Changed
- Preserve physical modifier keyCodes for standard Settings `Change...` recordings when the saved hotkey is modifier-only.
- Make the default dictation preset a shared `Fn` gesture: double-tap `Fn` for hands-free and hold `Fn` for push-to-talk.
- Migrate legacy default installs from `Fn+Space` hands-free plus `Fn` push-to-talk to the shared `Fn` gesture preset, while preserving explicit `Fn+Space` customizations when push-to-talk is custom.
- Keep `Fn+Space` available as a custom hands-free chord and fix recording it when macOS omits the function flag on Space keyDown.
- Allow only the built-in shared `Fn` default to overlap; custom overlapping dictation shortcuts remain blocked.
- Update hotkey specs, ADR notes, and manual QA for the new default preset.

## Root Cause
There were two hotkey modeling issues.

First, the standard modifier-chord recorder displayed physical left/right modifiers while recording, but persisted modifier-only chords without keyCodes. That made recordings like Left Option + Left Command behave as generic Option + Command at runtime.

Second, dictation defaults still modeled hands-free as `Fn+Space` even though the product default gesture is `Fn+Fn`. `Fn+Fn` should not be a customizable chord; it should be the built-in default gesture preset users can return to via restore default. Existing installs that had the old default pair persisted also needed migration so they do not remain stranded on `Fn+Space`. Separately, `Fn+Space` recording failed because the recorder only trusted keyDown modifier flags, and macOS can omit `.function` even after the recorder has observed Fn in pending modifier state.

## Verification
- `git diff --check`
- `swift test --filter HotkeyRecorderViewTests`
- `swift test --filter HotkeyManagerTests`
- `swift test --filter SideSpecific`
- `swift test --filter Hotkey`
- `swift test --filter SettingsHotkeyConflictMessageTests`
- `swift test --filter SettingsViewModelTests/testLegacyFnSpace --filter SettingsViewModelTests/testStoredFnSpaceHandsFreePersistsWhenDedicatedPushToTalkIsCustom --filter Hotkey`
- `swift test`

### DIFF
--- a/Sources/MacParakeet/App/AppHotkeyCoordinator.swift
+++ b/Sources/MacParakeet/App/AppHotkeyCoordinator.swift
@@ -116,6 +116,9 @@ final class AppHotkeyCoordinator {
         if handsFree.isDisabled && pushToTalk.isDisabled {
             return "Dictation Shortcuts: Disabled"
         }
+        if HotkeyTrigger.isDefaultDictationGesturePreset(handsFree: handsFree, pushToTalk: pushToTalk) {
+            return "Dictation: Hold Fn / Tap Fn"
+        }
         if handsFree.overlaps(with: pushToTalk) {
             let conflictName = handsFree == pushToTalk
                 ? handsFree.displayName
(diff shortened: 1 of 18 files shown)


moona3k:
Fresh review pass completed. Gemini and Greptile both reviewed the PR with no requested code changes; I also rechecked the recorder serialization path, runtime side-specific matcher, and duplicate-conflict validation against the regression. Waiting on CI before merge.

moona3k:
This branch picked up additional Fn-default work after the side-specific modifier PR was opened, so I split the regression fix into a clean narrow PR: #342. I am treating #342 as the merge target for the left/right modifier regression and leaving this broader branch untouched.

moona3k:
Update: after the latest direction, this PR is the merge target for the broader hotkey work: recorder persistence, Fn+Fn default preset, Fn+Space custom recording support, and legacy default migration. #342 was a temporary narrow split for the side-specific modifier regression and can be closed once this PR lands.

PR #342, opened 2026-05-23
Preserve side-specific modifier chords

moona3k:
## What Changed
- Preserve physical modifier keyCodes for standard Settings `Change...` recordings when the saved hotkey is modifier-only.
- Keep explicit either-side recording behavior unchanged; generic capture still strips physical sides by design.
- Add regression coverage for:
  - standard left/right modifier-only chords such as Left Option + Left Command,
  - left-shift + right-shift capture and runtime matching,
  - duplicate dictation-hotkey validation for side-specific Shift chords.

## Root Cause
The previous side-specific hotkey runtime fix was still intact. The regression was at the recorder persistence boundary: `HotkeyRecorderView.modifierChordTrigger` treated `.standard` recording the same as `.generic`, stripping each `ModifierComponent.keyCode` before saving.

That made the recording UI appear side-specific while the saved trigger became a generic modifier chord, so Left Option + Left Command could be triggered by either left or right modifiers.

## Verification
- `swift test --filter Hotkey`
- Prior local validation on the same narrow diff before the unrelated PR branch update: `swift test`, `swift test --filter HotkeyRecorderViewTests`, `swift test --filter HotkeyManagerTests`, `swift test --filter SideSpecific`, `swift test --filter SettingsHotkeyConflictMessageTests`

### DIFF
--- a/Sources/MacParakeet/Views/Settings/HotkeyRecorderView.swift
+++ b/Sources/MacParakeet/Views/Settings/HotkeyRecorderView.swift
@@ -377,8 +377,8 @@ struct HotkeyRecorderView: View {
     }
 
     /// Map a modifier name + physical keyCode to a bare modifier trigger.
-    /// Generic capture preserves the legacy "either side" behavior. Side-specific capture
-    /// records the physical left/right key that was used during recording.
+    /// Generic capture preserves the legacy "either side" behavior. Standard
+    /// and side-specific capture record the physical left/right key used.
     static func bareModifierTrigger(
         for name: String,
         keyCode: UInt16,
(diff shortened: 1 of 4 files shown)


moona3k:
Fresh review pass completed on the clean PR. Greptile reviewed with no requested changes, and I rechecked the recorder serialization path, runtime side-specific matcher, and duplicate-conflict validation against the regression. Waiting on CI before merge.

moona3k:
Gemini review checked as well: no feedback and no line comments. No Copilot comments are present. CI remains the only merge blocker.

PR #360, opened 2026-05-25
Subtitle polish: micro-cue fix, batched reviewer, Settings card

justingonz96-creator:
## What Changed

Four commits, all tightening the LLM-refined subtitle export path after the SRT (38) regression and the cost concerns flagged in the session that produced this branch.

### 1. `fix(subtitle): force-absorb micro-cues and auto-repair LLM index gaps` (43cb68de)

Root cause of the SRT (38) over-fragmentation: a model switch from `deepseek-v4-pro:cloud` to `gemma4:31b-cloud` exposed two new failure modes the existing pipeline didn't defend against:

- Gemma emits single-word cues whose audio duration is ~1 ms (Whisper word-token boundary). The 500 ms gap floor that protects normal-sized cues from being merged across utterance boundaries also protected these obvious fragments.
- Gemma occasionally drops a single word index when emitting JSON (`{end:9},{start:11}` — word 10 missing). `LayoutPlanParser` treated this as a hard failure, falling back entire 80-word chunks to deterministic layout.

Fixes:
- New `forceAbsorbMicroCues` pass (triggered on cue duration < 250 ms AND word count ≤ 3) that merges fragments into the better neighbour even across the 500 ms gap floor. Direction picking uses `.!?` cues to keep sentence integrity.
- New `autoCorrectGaps` in `LayoutPlanParser` that extends the previous range's end over a 1–3 word LLM gap. Wider gaps still fall back (silent absorption of 4+ words is worse than a chunk fallback).

Verified: SRT (38) 623 cues / 60+ sub-200 ms → SRT (39) 409 cues / 0 sub-200 ms on the same Gemma model.

### 2. `perf(subtitle): batch reviewer pairs to cut LLM call count ~5x` (93dafa11)

`SubtitleLLMReviewer` was issuing one LLM call per adjacent cue pair — ~400 calls per 30-min export. Now batches 5 pairs per call: 6-cue windows with numbered boundaries, response shape `{"decisions":[{"pair":0,"action":"keep"},...]}`.

New parser `ReviewActionParser.parseBatch` is defensive: a malformed individual decision is silently dropped (caller defaults to `.keep`), but a whole-batch JSON failure surfaces as a typed error. Accepts `index` as an alias for `pair` for model drift tolerance.

Wall-clock concurrency unchanged (still 4 batches in parallel); only call count drops. Progress reporting continues per pair so the UI bar stays smooth.

### 3. `fix(text): collapse Whisper apostrophe artifacts` (6b3c3fef)

SRT (39) cue 310 came back as `Y 'all feel that?` — Whisper's BPE tokenizer split `y'all` into two tokens with the apostrophe leading the second one. New `collapseWhisperApostropheArtifacts` runs alongside the existing hyphen-artifact pass and collapses the stray space when the right-hand side is a known contraction suffix (`ll`, `ve`, `re`, `s`, `t`, `d`, `m`, `all`). The `\b` anchor on the suffix prevents collapsing genuine quoted speech (`he said 'hello' loudly`).

### 4. `feat(settings): AI Subtitle Refinement card with batch-size knob` (80312acd)

New `SubtitleRefinementCard` on the Settings → AI tab. Surfaces the previously-buried `useLLMRefinement` toggle and exposes the new `reviewerPairsPerBatch` cost knob (slider 1...10, default 5). Persists via `TranscriptExportPreferences` so the existing Export popover toggle reads the same value. Shows the currently-effective LLM model name from `LLMSettingsViewModel` so the user can verify which model will field their refinement calls before committing to a 30-minute export.

## Root Intent

User reported "the splits are back" on an export after switching to Gemma 4 to reduce Ollama Cloud spend. That triggered:

1. Diagnose & fix the over-fragmentation (commits 1 + 3)
2. Cut the LLM cost so they don't have to use the cheap-but-noisy model for cost reasons (commit 2)
3. Make the feature discoverable and the cost lever adjustable without editing plist defaults (commit 4)

## Verification

- Full `swift test` clean across all four commits
- 11 new regression tests added across the touched modules:
  - 3 in `ApplyTimingPostProcessingTests` (micro-cue absorb backward, forward, and the standalone-short-sentence safety case)
  - 2 in `LayoutPlanParserTests` (narrow gap auto-repair, wide gap still falls back)
  - 1 in `SubtitleLLMLayoutPlannerTests` (end-to-end `testNarrowCoverageGapAutoRepairs`)
  - 3 in `SubtitleLLMReviewerTests` (batching, index mapping, missing-decisions fallback)
  - 9 positive + 2 negative apostrophe-artifact cases in `ApplyTimingPostProcessingTests`
  - 5 `parseBatch` cases in `ReviewActionParserTests`
- Generalization sanity check on a second, unrelated 20-min transcription (Eden Lusk Speed 20): 220 cues, 0 sub-200 ms cues, 0 apostrophe artifacts, 0 hyphen artifacts — the Marc Penna tuning did not overfit.

## ADRs Applied

None — this is iterative quality polish on existing functionality, not architecture. Touches:

- `Sources/MacParakeetCore/Services/ExportService.swift` and the subtitle pipeline it owns
- `Sources/MacParakeetCore/Services/Subtitle/` (reviewer, parser, layout planner)
- One new Settings view (`SubtitleRefinementCard.swift`)

No public API or schema changes outside the new `SubtitleExportConfig.reviewerPairsPerBatch` field (Codable backward-compatible via `decodeIfPresent ?? 5`).

## Test plan

- [ ] Open Settings → AI tab and confirm the new "AI Subtitle Refinement" card renders
- [ ] Toggle "Enable AI refinement..." off and confirm the slider hides
- [ ] Move the slider; reopen the Export Options popover and confirm the stored value persists
- [ ] Run an export on a 20–30 min transcription with AI refinement enabled; confirm cue count is healthy and no sub-200 ms cues appear


justingonz96-creator:
Closing — opened against the wrong target. Re-creating against justingonz96-creator/macparakeet.

PR #369, opened 2026-05-26
[codex] Fix dictation stop paste focus loss

### LINKED ISSUE #366: [Bug Report] When clicking to end the recording, it resets it and I lose the audio. I'm talki...
## Message

When clicking to end the recording, it resets it and I lose the audio. I'm talking about when I click stop and paste, it just loses the transcription.

## System Info

| Field | Value |
|-------|-------|
| App Version | 0.6.11 (20260524210220) |
| Git Commit | `d42138a00cdc` |
| macOS | 26.5.0 |
| Chip | Apple M5 |

---

moona3k:
## Summary
- keep the dictation overlay panel keyless so clicking Stop & paste does not disturb the user's focused input
- route persistent-mode Stop / Cancel mouse handling through AppKit hit zones instead of relying on a keyable SwiftUI button panel
- align mouse hit testing to tracker-local coordinates and fire control actions on mouse-up so drag-out cancels work
- add regression tests for keyless panel behavior, control-zone hit testing, coordinate conversion, mouse-up activation, and undersized bounds
- update the specs for the new pass-through-except-controls tracking behavior

## Root Cause
Commit `57ce8043` made the dictation overlay panel keyable so SwiftUI buttons inside the non-activating panel could receive clicks. That solved the original button clickability issue, but it meant clicking Stop & paste could make the overlay the key window before paste dispatch. The transcript still reached history, but the generated Cmd+V could miss the user's original focused input.

This keeps the original intent, clickable persistent-mode controls, while restoring the overlay's keyless contract.

## Review Feedback
- Addressed Gemini coordinate-space feedback by converting `hitTest(_:)` points to local tracking-view coordinates and evaluating closures against `tracker.bounds`.
- Addressed Gemini defensive bounds feedback by rejecting hit testing when the panel is narrower than the pill geometry.
- Addressed Greptile mouse-down semantics by moving activation to mouse-up with down/up hit guards.
- Addressed Greptile stale-comment note on `resignKeyWindow`.

## Validation
- `git diff --check`
- `swift test --filter DictationOverlayControllerTests` -> 8 tests, 0 failures
- `swift test` -> 3,002 XCTest cases, 9 skipped, 0 failures; 16 Swift Testing cases, 0 failures

Fixes #366

### DIFF
--- a/Sources/MacParakeet/Views/Dictation/DictationOverlayController.swift
+++ b/Sources/MacParakeet/Views/Dictation/DictationOverlayController.swift
@@ -7,10 +7,14 @@ import SwiftUI
 /// NSView overlay that detects mouse hover and position via NSTrackingArea with `.activeAlways`.
 /// Required because `.help()`, `.onHover`, and standard tracking options
 /// all fail on non-activating NSPanel. See CLAUDE.md Known Pitfalls.
-private final class MouseTrackingView: NSView {
+final class MouseTrackingView: NSView {
     var onEnter: (() -> Void)?
     var onExit: (() -> Void)?
     var onMoved: ((NSPoint) -> Void)?
+    var shouldReceiveMouseDown: ((NSPoint) -> Bool)?
+    var onClick: ((NSPoint) -> Void)?
+
+    private var mouseDownPoint: NSPoint?
 
     override func updateTrackingAreas() {
         super.updateTrackingAreas()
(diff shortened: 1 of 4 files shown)


moona3k:
Closing this for now rather than merging a speculative fix. Issue #366 has not confirmed the key discriminator yet: whether the transcript is actually missing or whether transcription succeeded and only auto-paste missed the target app.\n\nThis PR may be addressing a real focus-theft risk from the keyable dictation overlay, but that root cause is not verified for #366. The implementation also changes the overlay click contract enough that it should not land without covering the full set of overlay affordances, including cancel-window Undo/Dismiss behavior, not just recording-state Stop/Cancel.\n\nKeeping #366 open for more repro details is the cleaner path. If we later confirm a paste-focus failure, we can bring back a narrower hardening patch with full overlay hit-testing coverage.

PR #371, opened 2026-05-26
fix(text-processing): guard verb cardinal pass against prose false positive

justingonz96-creator:
## Summary

- `applyVerbCardinalAtEndPass` was firing on general prose: input `"I have four. Three of them are nice."` was getting converted to `"I have 4. Three of them are nice."`. The pass exists to catch cross-cue countdown remnants in fitness instruction (`"back and do two"` / `"we do it two"` where the unit `"rounds"`/`"times"` is in the next cue), but the regex alone couldn't tell those apart from ordinary prose with the same shape.
- Adds a has-digit guard mirroring the existing [`applyBareSpelledAtCueEndPass`](https://github.com/justingonz96-creator/macparakeet/blob/main/Sources/MacParakeetCore/TextProcessing/NumberNormalizer.swift#L442-L451) pattern: only fire when the cue already contains a `\b[1-9]\b` digit cardinal somewhere. The real cross-cue countdown remnant always has a sibling digit in the same cue; prose doesn't.
- Pre-existing bug surfaced during [PR #4](https://github.com/justingonz96-creator/macparakeet/pull/4) review. The period-countdown pass that PR added is correctly anchored at cue start and was never the trigger — the verb pass had the same general-prose shape problem all along.

## Test plan

- [x] New `testVerbCardinalAtEndDoesNotConvertProse` fails before the fix (`"I have 4. Three of them are nice."` vs expected `"I have four. Three of them are nice."`)
- [x] New test passes after the fix
- [x] `swift test --filter NumberNormalizer` — 29 tests, all pass (no regressions on legitimate fitness-context matches)
- [x] `swift test` — full suite 3206 tests pass, 10 skipped, 0 failures


justingonz96-creator:
Mistargeted — this fork ships from justingonz96-creator/macparakeet. Reopening on the right repo.

PR #386, opened 2026-05-29
Harden Meetings page for release: recurring-meeting preview fix + Ask Prompts copy

moona3k:
Pre-release hardening of the new Meetings page. Two independent changes:

## 1. Fix recurring-meeting collapse in the Upcoming preview 🔴
`CalendarEvent.id` is EventKit's `eventIdentifier`, which is **shared across every occurrence of a recurring series** — the identity note in `CalendarEvent.swift` explicitly warns that `id` is not occurrence-unique and that occurrence-level work must key on `dedupeKey`.

The new Meetings UI renders `ForEach(viewModel.upcomingEvents)` keyed on `id` (plus an `event.id != last?.id` hairline check). A user with a **daily standup** gets multiple occurrences sharing one `id` inside the 7-day fetch window → SwiftUI duplicate-ID "undefined results" (collapsed / mis-diffed rows) in the preview. The existing test mock never caught it because every `makeEvent` got a fresh UUID.

**Fix:** `MeetingsWorkspaceViewModel` now collapses occurrences to the **soonest per series** before capping to `upcomingEventLimit` (new `collapseRecurringOccurrences` helper). Display-only — the coordinator still acts on every occurrence via `dedupeKey`. Bonus UX: the preview shows distinct meetings instead of the same standup four times. New regression test `testUpcomingPreviewCollapsesRecurringOccurrencesToSoonest`.

## 2. Audience-friendly Ask Prompts copy ✍️
Rewrote the five instruction strings in `AskPromptsSheet` to drop internal jargon ("surface as quick pills", "after-response strip", "empty Ask state", "sparkle menu", "group label clusters") for plain language. References the user-visible ✨ menu and Ask tab directly.

## Testing
- `swift test`: **3037 tests, 0 failures, 9 skipped**
- `swift build --target MacParakeet`: clean (copy is app-target only)

### DIFF
--- a/Sources/MacParakeet/Views/MeetingRecording/AskPromptsSheet.swift
+++ b/Sources/MacParakeet/Views/MeetingRecording/AskPromptsSheet.swift
@@ -38,7 +38,7 @@ struct AskPromptsSheet: View {
                     zone(
                         title: "Pinned",
                         countSuffix: "\(viewModel.pinnedCount)",
-                        subtitle: "Your most-used questions, shown as one-tap buttons during a live meeting. The order here is the order you'll see them in.",
+                        subtitle: "Your most-used questions, shown as quick buttons during a live meeting. The order here is the order you'll see them in.",
                         rows: viewModel.allPinned,
                         pinned: true
                     )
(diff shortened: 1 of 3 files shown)


moona3k:
Addressed both reviewers in `74a…` (latest push):
- **Gemini** (line 231 questions/prompts inconsistency): fixed — both ternary branches now use "questions."
- **Greptile** ("tap" vs macOS "click" convention): confirmed against the codebase — the app reserves "Tap"/"Double-tap" for the hands-free *hotkey* press and uses "Click" for on-screen UI element actions ("Click the dictation pill", "Click to rename"). Switched the literal instruction to "Click the pin icon" and replaced the "tap" metaphors with neutral wording ("front and center", "quick buttons", "handy"). Good catch.

PR #388, opened 2026-05-29
Surface post-meeting auto-notes on the Meetings page (source-scoped auto-run)

moona3k:
Surfaces the highest-value meeting feature — automatic post-meeting notes (Summary, Action Items, …) — on the Meetings page, and makes prompt auto-run **source-scoped** so it's an honest meeting-only promise.

## Why
Post-meeting auto-run already worked, but it was (1) **invisible** — only reachable via a meeting detail → `+` → Manage Prompts, and (2) **global** — the same auto-run fired on file/YouTube transcriptions, with no source scoping. That global leak is exactly what forced the Memo-Steered Notes revert (ADR-020). This implements path (a) from the ADR-020 amendment.

## What's in it
**Source-scoped auto-run (infra)**
- `Prompt.appliesToSources: Set<Transcription.SourceType>?` — `nil` = all sources (= historical behavior; every existing/migrated row decodes to nil, so **zero regression** for users who never touch the new card). Rule centralized in `Prompt.autoRuns(for:)`.
- Migration `v0.20-prompt-applies-to-sources`: nullable JSON column only — no row inserts, so the historical Codable-seed trap is avoided; the v0.7 seed is untouched and the reconciler preserves the field on built-ins.
- `PromptRepository.fetchAutoRunPrompts(for:)` (source filter) + `setAutoRun(id:source:enabled:)` (per-source transitions that never leak onto unintended sources). Global `toggleAutoRun` clears scope on enable ("all sources" = the reset path).
- Trigger threads `transcription.sourceType` → `PromptResultsViewModel.autoGeneratePromptResults`.

**Meetings "After Each Meeting" card (UX)**
- New card between Intelligence and Live Ask: toggle chips for each result prompt (scoped to `.meeting`), a "Uses {provider}" line, and a **Manage** deep-link to the full Prompt Library. Gated on a configured AI provider (shows "Set up AI" instead of dead toggles).

## Testing
- `swift test`: **full suite green, 0 failures** (9 new tests: source-scoped query, all `setAutoRun` transitions, global-toggle reset, reconciler preservation, source-aware trigger, Meetings VM exposure/toggle).
- `swift build --target MacParakeet`: clean.
- Two independent fresh-eye reviews (Codex + Gemini): both **merge-ready, no blockers**.

## ADRs
ADR-020 (2026-05 amendment) — marks path (a) `appliesToSources` as shipped. ADR-013 — preserves "zero auto-run is valid."


moona3k:
Addressed all three review comments in `a76da01d`:
- **Scope normalization (Greptile P2):** `setAutoRun` now collapses a set that covers every source back to `nil` after enabling, so the canonical "all sources" form is preserved and a future `SourceType` case is auto-included rather than silently excluded. Mirrored in the mock + added `testSetAutoRunReEnablingEverySourceNormalizesScopeToNil`. Good catch.
- **"Tap" → "Click" (Greptile):** fixed in the card copy and chip help text — matches the macOS pointer convention the app uses for on-screen actions.
- **Redundant `.category == .result` (Gemini + Greptile):** removed; `promptsViewModel.prompts` is already result-only.

Full suite green; CI re-running on the new commit.

moona3k:
## Fresh-eye review round 2 (Codex + Gemini)

Both reviewers returned the core diff merge-ready (no P0/P1 inside the diff). They surfaced two real gaps *outside* the diff, plus nits — all addressed in `639f9ee5`:

- **CLI `prompts set` scope leak (P1, real):** `--auto-run`/`--no-auto-run`/`--hidden` mutated the struct directly and bypassed the `appliesToSources = nil` normalization that the GUI's global toggle enforces. A prompt narrowed to meetings-only in the GUI, then `--auto-run` via CLI, would claim global-on while silently staying meeting-scoped — the exact leak this feature prevents. Fixed; extracted the mutation into a pure `SetSubcommand.applyFlags(...)` and added 3 CLI tests.
- **`restoreDefaults()` left scoping (P2):** now resets `appliesToSources = NULL` so built-ins return to shipped unscoped state. Added repo test.
- **Spec drift (P2):** `spec/01-data-model.md` updated — schema block, Swift struct, migration list, notes entry.
- **Nits:** hoisted `isOn` in `AutoNoteChip` (single read for render + tap); added partial-scope-add repo test.

Confirmed clean by both: state-machine transitions, idempotent v0.20 migration, NULL→nil GRDB round-trip, single auto-run call site threads `sourceType`, reconciler preserves scoping. Focused suites green (PromptRepository, PromptsCommand, Meetings/PromptResults/Prompts VMs).

PR #445, opened 2026-06-06
[codex] Add multi-select dictation cleanup

### LINKED ISSUE #425: [Feature Request] Allow multi-select when cleaning up dictations
## Message

Allow multi-select when cleaning up dictations

## System Info

| Field | Value |
|-------|-------|
| App Version | 0.6.17 (20260601201518) |
| Git Commit | `20f4daac5e68` |
| macOS | 26.5.0 |
| Chip | Apple M4 Pro |

---

moona3k:
## Summary

Adds multi-select cleanup to Dictation History so users can select several visible dictations and delete them with one confirmation instead of deleting rows one at a time.

## User Flow

```text
History list
    |
    +-- select one or more visible rows
            |
            v
    selection bar appears
    [N selected] [Select All] [Clear] [Delete]
            |
            v
    confirmation dialog
            |
            v
    selected history rows and retained audio files are deleted
```

## What Changed

- Adds row selection toggles and selected-row styling in Dictation History.
- Adds a compact bulk-action bar for select all, clear, and delete.
- Keeps selection scoped to the visible history/search results, while pending delete confirmation stores row snapshots so the delete target is stable.
- Reuses the existing delete semantics for single and batch cleanup, including audio-file removal, telemetry, playback stop, and one history refresh after completion.
- Moves the disk/database delete loop off the main actor so larger cleanup actions do not block the UI.
- Updates the feature spec and traceability docs for the new cleanup behavior.

## Verification

- `swift test --filter DictationHistoryViewModelTests`
- `swift test`
- `git diff --check`

Closes #425

### DIFF
--- a/Sources/MacParakeet/Views/History/DictationHistoryView.swift
+++ b/Sources/MacParakeet/Views/History/DictationHistoryView.swift
@@ -37,20 +37,20 @@ struct DictationHistoryView: View {
         .animation(DesignSystem.Animation.contentSwap, value: viewModel.playbackError != nil)
         .animation(DesignSystem.Animation.contentSwap, value: viewModel.selectedSubTab)
         .alert(
-            "Delete Dictation?",
+            deleteAlertTitle,
             isPresented: Binding(
-                get: { viewModel.pendingDeleteDictation != nil },
-                set: { if !$0 { viewModel.pendingDeleteDictation = nil } }
+                get: { viewModel.pendingDeleteCount > 0 },
+                set: { if !$0 { viewModel.cancelPendingDelete() } }
             )
         ) {
             Button("Cancel", role: .cancel) {
-                viewModel.pendingDeleteDictation = nil
+                viewModel.cancelPendingDelete()
             }
             Button("Delete", role: .destructive) {
-                viewModel.confirmDelete()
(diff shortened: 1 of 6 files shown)


moona3k:
Addressed the review feedback in `a88fd913`:

- Batch delete now stores pending `Dictation` snapshots instead of only IDs, so the confirmed delete target stays stable across search/reload changes.
- The disk/database delete loop now runs off the main actor and refreshes history once after completion.
- The selected-actions bar now has an explicit animation trigger for the transition.
- Added/updated History view-model tests for the async delete path and pending snapshot behavior.

Validation:
- `swift test --filter DictationHistoryViewModelTests`
- `swift test`
- `git diff --check`

moona3k:
You're right to flag this: Greptile posted after the PR was already merged, so those two comments were not included in the pre-merge fix pass.

Fixed directly on `main` in `3a344c45`:
- Cached the delete-alert count while the alert is shown, so dismissing a multi-delete confirmation no longer flashes to the singular copy.
- Replaced the tight 1s async polling helper with an `XCTestExpectation`-backed wait with a 5s timeout.

Validation:
- `swift test --filter DictationHistoryViewModelTests`
- `swift test`
- `git diff --check`

PR #465, opened 2026-06-08
Fix Decide Transform default hotkey

### LINKED ISSUE #453: [Feedback] ⌥3 is not a great choice for the default transform hotkey as it is needed for # ...
## Message

⌥3 is not a great choice for the default transform hotkey as it is needed for # on MBP. (I’ve reset to ^⌥3 which might be a better starting point for the transforms)

## System Info

| Field | Value |
|-------|-------|
| App Version | 0.6.21 (20260607023821) |
| Git Commit | `3c7f3d8f82ce` |
| macOS | 26.5.1 |
| Chip | Apple M3 Max |

---

moona3k:
## Summary
- change the built-in Decide Transform default from `Option-3` to `Control-Option-3`
- migrate existing built-in Decide rows still using the old default while preserving cleared/custom shortcuts
- update current-facing Transform docs/spec copy and regression coverage

## Root Cause
`Option-3` is needed for `#` on some MacBook keyboard layouts, so using it as a global default Transform hotkey steals a normal typing chord.

## Verification
- `git diff --check`
- `swift test --filter KeyboardShortcutTests --filter PromptRepositoryTests --filter TransformsViewModelTests`
- `swift test`

Closes #453


moona3k:
Addressed Greptile’s doc-comment finding in 3f0824292. Re-ran `git diff --check`, the focused shortcut/repository/view-model tests, and the full local `swift test` suite on the updated head.

PR #472, opened 2026-06-09
URL card: name matched platform + enrich link-failure detail + platform telemetry

moona3k:
**Review pass for two changes already on `main`** (`bc64a2ed3`, `c8b84256d`). Base is pinned just before them so the diff is reviewable. Findings will be addressed fix-forward on `main`; this PR won't merge into `main`.

## 1. Dynamic URL-card title (`bc64a2ed3`)
The Transcribe card + floating input panel headings now name the recognized platform ("Transcribe TikTok" / "Transcribe Vimeo" / …) instead of a static "Transcribe YouTube & more", so glyph + title + caption move together. Falls back to the generic title while idle / on an unrecognized link. Cross-fades to match each surface's existing glyph animation.

## 2. Richer link-failure diagnostic (`bc64a2ed3`)
A failed URL download now produces a copyable block (`errorDetail`) carrying the source link + app/macOS/chip, distinct from the telemetry-classified `errorMessage` (so the URL never enters telemetry). URL lane only; file/batch failures keep the headline.

## 3. `platform` telemetry sub-dimension (`c8b84256d`)
Adds an additive `platform` prop (youtube · x · vimeo · … · apple_podcasts · … · other) on the `transcription_operation` event, sourced from `MediaPlatform.recognize()`, without renaming `source` (no historical-series break). Set on the two URL lanes; absent on file/meeting. Non-identifying enum only — never the raw host.

## Tests
Full suite green (build clean). New: URL-failure diagnostic format + file-failure-leaves-nil; telemetry platform serialization + file-source-omits-platform + enum bucketing (apple_podcasts snake_case, nil → other).


moona3k:
All review feedback addressed (pushed to `main` + this branch):

- **🔴 Stale `errorDetail` (Gemini, HIGH)** — fixed exactly as suggested: `errorMessage` now has a `didSet { errorDetail = nil }`, centralizing the invariant so any headline change drops a diagnostic built for a prior failure (covers the `unsupportedDropMessage`, deletion, and missing-config paths). Removed the now-redundant explicit clears. Added a regression test: a URL failure followed by an unsupported-file drop leaves `errorDetail == nil`. (`8a3e3a4fb`)
- **🟡 `urlCardTitle` / `headerTitle` → `LocalizedStringKey` (Gemini, MEDIUM ×2)** — done. The original static title was a string literal (localizable); the computed `String` had regressed that. (`7d9625020`)
- **🟡 Panel title animation keyed on `headerTitle` (Greptile, P2)** — done; now matches `TranscribeView` and is the more direct signal for `contentTransition`. (`7d9625020`)

Full suite green, build clean. (Note: these changes already shipped to `main`; this PR is the review gate, so it won't merge into `main`.)

moona3k:
@greptileai review — all four findings (stale errorDetail HIGH, LocalizedStringKey ×2, animation keying P2) addressed in 8a3e3a4fb + 7d9625020. Please re-review the updated branch.

moona3k:
Addressed in `8e45afd34`: both title animations now key on the `MediaPlatform?` enum (`recognizedURLPlatform` / `draftPlatform`) rather than the `LocalizedStringKey` title.

Minor note for the record: it actually compiles fine as-is — `LocalizedStringKey` *does* conform to `Equatable` (the build was green) — so it wasn't a compiler error. But the suggestion is the right call regardless: the enum is a reliable change-signal (LSK equality is opaque/interpolated), and keying both surfaces on the platform also resolves the earlier consistency note (Greptile P2) and keeps the panel title in sync with its glyph.

/gemini review

moona3k:
Final cleanup pushed (`efed51ae8`): the two new error-diagnostic tests now use the file's `waitUntil { !self.viewModel.isTranscribing }` helper instead of a fixed `Task.sleep` — deterministic and ~0.03s vs 200ms each. That closes the last two re-review nits.

Net state: Greptile **5/5** on the latest commit, all inline findings across both bots addressed (stale-`errorDetail` → `setError`/`clearError` choke point, `LocalizedStringKey` titles, enum-keyed animations, `waitUntil` tests). Full suite green. All changes are live on `main`.

moona3k:
Closing this review gate — its purpose is served. Greptile **5/5**, all findings from both bots addressed, and every commit is already live on `main` (`efed51ae8`). This PR's base was a throwaway pin before the batch, so there's nothing to merge into `main`. Deleting the temporary `review*/url-card-platform` branches.

PR #492, opened 2026-06-11
Stop recording phantom fn on function-family hotkeys

athurdekoos:
## Summary

The Settings hotkey recorder mistakes macOS's synthetic `.function` modifier flag for a held Fn key. macOS sets `.function` on **every** function-family key press (F1–F20, arrows, nav cluster) even when the physical Fn/globe key is untouched, so:

- Recording **Ctrl+F19** captures **`fn+⌃+F19`** instead of **`⌃F19`** — wrong label, and the polluted trigger demands `NX_SECONDARYFNMASK` at runtime, which synthesized F-key events (QMK / Karabiner / programmable keyboards) often omit → flaky or dead hotkey.
- A **bare F-key can never record as a standalone key**: the phantom fn makes `heldModifiers` non-empty, so `HotkeyRecorderView`'s bare-key path is unreachable for the entire function-key family.

Found while a user with a programmable keyboard tried to bind Ctrl+F19 push-to-talk / double-tap-⌃F19 hands-free (the shared `.doubleTapAndHold` gesture). The recorder bug made that configuration impossible to enter.

## What Changed

- **`HotkeyRecorderView.chordModifierNames`** now takes the pressed keyCode and only derives `"fn"` from the `.function` flag when the key is **not** function-family. A genuinely held Fn still arrives via `pendingModifierComponents` (flagsChanged keyCode 63/179) and is unioned in unconditionally; fn+letter chords keep flags-derived fn (letters aren't function-family).
- **`KeyCodeNames.isFunctionFamilyKeyCode(_:)`** — new predicate (F1–F20, arrows, nav cluster, Help) colocated with the display-name table.
- **`HotkeyTrigger.ignoredChordEventFlags` + `GlobalShortcutManager`** — exact chord matching now masks out the phantom fn bit for function-family chord keys whose trigger doesn't itself require fn. Without this, the recorder fix would regress *hardware* F-key chords for meeting/file/YouTube shortcuts (their newly-clean triggers would stop exact-matching hardware events that do carry the bit). Triggers that genuinely include `"fn"` keep their exact requirement, so previously persisted polluted triggers behave exactly as before.
- **`HotkeyManager` (dictation): no change needed** — its superset chord match already tolerates the bit either way; pinned with a characterization test.

## Tests

Red-first: `testFunctionKeyChordTriggersWithPhantomFnFlagBit` (GlobalShortcutManager exact match), `testChordCaptureIgnoresFunctionFlagForFunctionKeyChord` / `testBareFunctionKeyCaptureIgnoresFunctionFlag` / `testChordCaptureIgnoresFunctionFlagForArrowAndNavKeys` (recorder) all failed against the old code — the recorder failure reproduces the user-visible symptom verbatim (`"fn+⌃+F19"` ≠ `"⌃F19"`). Preserved-behavior pins: pending-fn chords, fn+letter via flags, fn-requiring chords still need the fn bit, HotkeyManager superset tolerance. Full `swift test` green.

Manual: with both dictation shortcuts recorded as ⌃F19, Settings shows **Hold ⌃F19 / Double-tap ⌃F19** and both gestures work (hardware and synthetic F19, with/without the fn bit).

## Out of Scope

- **Migration of previously persisted polluted triggers** — `HotkeyManager`'s superset matching keeps hardware users working, and auto-stripping fn is indistinguishable from a genuine fn+F-key chord intent. Re-recording yields the clean form.
- **`TransformsHotkeyRegistry`** — separate capture path that never records fn; its exact dispatch table has a pre-existing sibling issue for F-key Transform bindings (`relevantModifierBits` includes the fn bit). Happy to file a follow-up issue.

ADR-009 applied: matching stays masked to the five relevant modifier bits; this narrows when the fn bit counts as a *held modifier* versus key-event noise. REQ-DICT-004 traceability row updated (`KeyCodeNames.swift` / `KeyCodeNamesTests.swift`).


moona3k:
LGTM & thank you for the contribution :)

(this will be included for the next/future release)

Issue #478, opened 2026-06-10
[BUG] Summary generation with Codex doesn't work


andreagrandi:
I've configured MacParakeet to use Codex for external processing, as you can see here:

<img width="769" height="535" alt="Image" src="https://github.com/user-attachments/assets/f307a796-6097-46be-8dbe-8382004a091c" />

After I finish recording a meeting I ask the app to produce a Summary. It starts doing this:

<img width="767" height="412" alt="Image" src="https://github.com/user-attachments/assets/1083e896-77fa-4b72-8507-0a945d63d4b5" />

and then it goes back to this:

<img width="760" height="492" alt="Image" src="https://github.com/user-attachments/assets/9bcd3ca2-01d9-4dc2-883f-e9fdb1a8bd4f" />

I'm using this version:

MacParakeet Build Identity
Version: 0.6.21
Build: 20260607023821
Source: dist-xcodebuild-release
Commit: 3c7f3d8f82ce
Built: 2026-06-07T02:38:27Z
Executable: /Applications/MacParakeet.app/Contents/MacOS/MacParakeet
Bundle: /Applications/MacParakeet.app

Is this a known bug? Thanks

moona3k:
This looks like a bug. Let me try to repro and get this addressed. Thanks for raising.

moona3k:
@andreagrandi So the LLM prompt generation was only covering the happy path. In the event of a failed call, the tab disappeared and there was no UI for recovery/retry. Well, we now have a UI (along with the error details).

<img width="1414" height="956" alt="Image" src="https://github.com/user-attachments/assets/1000e52b-8242-49f0-8a65-3bc3e3932f38" />

As for why the codex call failed in your case.. it's likely due to the conservative default timeout, which is 45s; it seems like your transcript processing had exceeded the timeout window. In any case, I've bumped up the default to 300s (and existing configs on the previous 45s default are migrated automatically).

This will be included in the next release. Until then... please try bumping up the timeout to something longer like 300s. 
Now, if it still fails after that, please run the below + share/paste the output for further debugging. 

```
log show --last 2h --predicate 'subsystem == "com.macparakeet.viewmodels"' | grep "Failed to generate"
```

andreagrandi:
> [@andreagrandi](https://github.com/andreagrandi) So the LLM prompt generation was only covering the happy path. In the event of a failed call, the tab disappeared and there was no UI for recovery/retry. Well, we now have a UI (along with the error details).
> 
[...]

PR #494, opened 2026-06-11
Give the transcripts AI Formatter a Settings toggle and a length cap

### LINKED ISSUE #493: AI Formatter for transcripts has no UI toggle and stalls meeting finalization for the full timeout on long transcripts
## Summary

The AI Formatter for **file and meeting transcripts** is force-enabled with no UI control, and it runs **inline** in transcription finalization. For meeting-length transcripts the formatter task (full-text rewrite) cannot complete within the CLI/provider timeout, so every meeting burns the entire timeout (300s since #478), fails, and falls back to standard cleanup — delaying the transcript's availability by ~5 minutes for zero benefit.

## Details

**1. No user control over the transcripts-side formatter.**
`LLMSettingsViewModel.persistAIFormatterPreferences` sets `aiFormatterEnabled = (draft.providerID != nil)` on every settings save. Saving any LLM provider silently turns the formatter on for file/meeting transcripts. The only visible switch in Settings → AI is "Use for dictation" (`aiFormatterEnabledForDictation`). There is no way to turn the transcripts pass off from the app, and any `defaults write` workaround is reverted on the next settings save.

**2. The formatter blocks finalization and can never succeed on long transcripts.**
`TranscriptionService.completeTranscription` awaits `formatTranscriptIfNeeded` over the full transcript before saving the record as completed. The formatter prompt requires reproducing the entire transcript ("Do not summarize… output only the final cleaned text"), i.e. output ≈ input length. For a 60–95 min meeting (~44k–66k chars ≈ 12–17k output tokens) this cannot finish within the timeout on a Local CLI provider.

Evidence from my `llm_runs` table (Local CLI / `codex exec --model gpt-5.4-mini`): the only two `formatter_transcription` runs on record both failed at exactly the timeout cap — 45.2s (pre-#478 default) and 300.2s (current default). 100% failure rate on meetings, with the full timeout burned each time before the silent fallback.

For comparison, a meeting **summary** over the same transcript through the same provider completes in ~13–17s, so this is specific to the formatter's full-rewrite workload, not to the provider.

## Suggested fix

1. Add a visible "Use for file & meeting transcripts" toggle next to the existing dictation toggle, instead of force-enabling on provider save.
2. Skip the formatter above an input-length threshold (return the standard-cleanup text directly) rather than spending the full timeout on a call that cannot succeed.

I have a fix for both on a branch and will open a PR shortly.

## Environment

- main @ 30a1bc4b (dev build)
- LLM provider: Local CLI (`codex exec --skip-git-repo-check --ephemeral --model gpt-5.4-mini`, 300s timeout)
- Meetings affected: 60 min / 43,880 chars and 95 min / 66,276 chars

andreagrandi:
Fixes #493

## Problem

The AI Formatter for **file and meeting transcripts** is force-enabled whenever an LLM provider is saved, with no UI control — the only visible switch ("Use for dictation") covers a different path. `LLMSettingsViewModel.persistAIFormatterPreferences` sets `aiFormatterEnabled = (provider != nil)` on every settings save, and `AppEnvironment.syncAIFormatterAvailabilityWithLLMConfiguration` re-asserts it on every launch, so even a manual `defaults write` opt-out is silently reverted.

On meeting-length transcripts the formatter call can never succeed: the prompt requires reproducing the full text, so output length tracks input length (~12–17k tokens for an hour-long meeting), which doesn't fit inside realistic provider timeouts. Because `completeTranscription` awaits the formatter inline, every meeting stalled finalization for the **entire timeout** (300s since #478) before failing and falling back to standard cleanup.

Evidence from a live install (Local CLI / `codex exec --model gpt-5.4-mini`): the only two `formatter_transcription` rows in `llm_runs` both failed at exactly the timeout cap — 45.2s (pre-#478 default) and 300.2s — on 43,880- and 66,276-char meeting transcripts. A *summary* over the same transcripts through the same provider completes in ~13–17s, so the provider is fine; the formatter's full-rewrite workload is the problem.

## Fix

1. **"Use for transcripts" toggle** (Settings → AI, next to the dictation toggle). New `aiFormatterEnabledForTranscriptions` preference, default `true` so shipped behavior is unchanged; the transcription gate in `AppEnvironment` is now availability AND this flag, mirroring the #408 dictation gate. Save no longer clobbers the user's choice; clear restores the default.
2. **Input-length cap**: `formatTranscriptIfNeeded` skips inputs over `AIFormatter.maxTranscriptionInputChars` (20k chars) and falls through to standard cleanup with a structured log line, instead of burning the full timeout on a call that cannot finish. This protects hour-long meetings even with the toggle on.
3. **Spec corrections**: `spec/02-features.md` (F8) and `spec/11-llm-integration.md` claimed meeting transcripts bypass the AI Formatter — they don't (meeting finalization shares `completeTranscription`, confirmed by the `llm_runs` rows above). Both passages now describe the real call path, the toggle, and the cap. Added REQ-LLM-005 + traceability row.

## Testing

- `swift test`: 3,578 tests, 0 failures.
- New coverage: preference default + persistence round-trips, transcription gate conjunction (including unset-key → pre-#493 behavior), save preserves an existing opt-out, clear restores the default, and over-cap transcripts never call the formatter and write no `llm_runs` row.
- Note: one pre-existing `DictationServiceTests` telemetry flake (races on the global `Telemetry` sink) failed once in the baseline run on `main` and did not reproduce; unrelated to this change.


moona3k (approved):
Superseded after the branch was rebased onto current main. Current approval is below.

moona3k (approved):
LGTM. Scoped fix: transcript formatting gets its own switch, opt-outs survive save and launch, and long transcripts skip LLM formatting before timeout risk. Full local Swift suite and hosted swift-test are green on the current head.

moona3k:
LGTM, thank you for the fix!

andreagrandi:
Thanks to you for accepting it! I've a couple of more ideas for improvements. I will try to contribute them in the next days.

PR #519, opened 2026-06-14
Start the speech-model download at onboarding open (dictation-first Part B)

moona3k:
## What & why

Part B of the dictation-first onboarding plan (`plans/active/2026-05-dictation-first-onboarding.md` §5). Part A (#515) trimmed onboarding to 6 steps; **Part B starts the ~465 MB Parakeet speech-model download when onboarding _opens_** instead of when the user reaches the second-to-last Speech Model step, so the download overlaps the Microphone / Accessibility / Hotkey steps. Timing change, not contents — on a fast connection the engine step is already `.ready` and the user never waits at a progress bar.

## The four guards (plan §5)

| Guard | What |
|---|---|
| §5.1 Decouple busy | Warm-up tracks its own `engineBusy`, separate from the permission `isBusy`, so the head-start download never disables the Microphone/Accessibility grant buttons. No view changes — the engine step renders from `engineState`. |
| §5.2 CJK fork | `whisperRecommendation` resolves synchronously in `init`, so the early trigger honors the Whisper path for CJK locales. |
| §5.3 Suppress early failures | Every failure routes through `applyEngineWarmUpFailure`, which surfaces `.failed` only when `step == .engine`; before that it resets to `.idle` so the engine step's `.onAppear` retries cleanly (no failed-card flash on an earlier step). Gating on `step` keeps every existing failure test green. |
| §5.4 Telemetry | `modelDownloadStarted` now fires at onboarding open; the start→ready duration still measures real download time (the background download runs independent of the user's step). |

The early trigger lives in `OnboardingFlowView`'s top-level `.onAppear`; the engine step's existing `.onAppear { startEngineWarmUp() }` stays as an idempotent fallback (generation + observation-token guards make the second call a no-op).

Also silences a Swift-5-mode concurrency warning on the static warm-up timeout (`nonisolated`; Swift 6 mode already treats it that way via SE-0434).

## Tests
- `testHeadStartWarmUpDoesNotHoldPermissionIsBusy` (§5.1)
- `testEngineStepRetriggerAfterHeadStartDoesNotDoubleDownload` (idempotency)
- `testHeadStartTakesWhisperPathForCJKLocaleBeforeEngineStep` (§5.2)
- `testWarmUpFailureBeforeEngineStepIsSuppressedUntilEngineStep` (§5.3) — **mutation-verified** (breaking the `step` gate fails this test)
- Existing stall test updated to assert `engineBusy` (the decoupled flag).

Full suite green (3696 tests, 0 failures); Swift-6 language-mode build clean.

## Risk
MED (perturbs the warm-up state machine) — mitigated: one early-trigger line in the view, the failure gate is transparent to all existing failure tests, and `startEngineWarmUp` idempotency is the same property the v0.4.22 generation guards were built for.


moona3k:
## Review resolution

Two independent fresh-eye reviews; both findings fixed and **mutation-verified**:

- **Auto-retry from `.failed` (engine-step boundary race)** → `036facbc5`: added a `.failed` guard to `startEngineWarmUp()` so a head-start failure that surfaces just as the engine step appears shows the Retry button instead of silently re-attempting. Only `retryEngineWarmUp()` (which resets to `.idle` first) restarts. New test `testStartEngineWarmUpDoesNotAutoRestartFromFailedState` (fails if the guard is removed).
- **Idempotency test masked by the mock's own dedup** → `036facbc5`: it asserted `warmUpCallCount == 1`, which `MockSTTClient.backgroundWarmUp()`'s internal `backgroundWarmUpTask != nil` guard keeps at 1 regardless of the VM guard. Added `backgroundWarmUpCallCount` (incremented before the mock's dedup) and assert on it — removing the VM's `warmUpObserverTask != nil` guard now fails the test.
- **Telemetry**: documented that `modelDownloadStarted/Failed` are attempt-counts (a suppressed pre-engine failure silently re-attempts at the engine step); `modelDownloadCompleted` still fires once.

Deliberately out of scope (confirmed by review):
- The **Whisper path has no stall watchdog** — pre-existing (it never did, even before Part B); a CJK head-start that hangs has the same exposure it had at the engine step. Worth a follow-up to extend the watchdog (and its test) to the Whisper path.
- **Back is enabled during the engine download** — an intentional side-effect of decoupling `isBusy`; arguably better UX (warm-up runs in the background, re-entry is idempotent).

Full suite green (**3697 tests, 0 failures**); Swift-6 language-mode build clean.

moona3k:
## Review pass (Claude) — one fix pushed, findings adjudicated

Reviewed the warm-up state machine end-to-end and ran an independent second-opinion pass. **Pushed `9acb60f05`** addressing the two open bot findings, plus an adjudication of the rest so the deferred items are explicit.

### Fixed
- **`engineBusy` leak on cancel (Gemini, HIGH).** `cancelWarmUpObservation()` never cleared `engineBusy`, so a window close mid-warm-up (`stopObservingWarmUp()`) left it `true` forever — the cancelled observer task's `defer { clearObservationIfCurrent }` bails on the just-nilled `warmUpObservationToken` and never reaches a busy-clearing terminal state. Cleared it in the teardown so the invariant holds across all three callers (window-close, stall watchdog, retry). Currently latent (no surface binds to `OnboardingViewModel.engineBusy` yet) but a real break in the state machine Part B perturbs. New **mutation-verified** test `testCancellingWarmUpObservationClearsEngineBusy` (fails if the reset is removed).
- **Duplicate "Guard 2" test label (Greptile, P2).** Relabeled the idempotency test to `Guard 1 / idempotency (§5.1)` so guard numbers map 1:1 to plan §5.x (§5.1→1, §5.2→2, §5.3→3).

### Adjudicated — deliberately out of scope (pre-existing, in the Whisper path you already flagged for follow-up)
- **Whisper warm-up has no stall watchdog** — exactly as your review note says; pre-existing, Part B only widens the window. A hung CJK download has the same terminal exposure at the engine step it had before. Follow-up.
- **Whisper cancel can persist the engine preference** — `activateWhisperEngine`'s `SpeechEnginePreference.whisper.save(...)` isn't generation-guarded, so a cancel landing during activation can still persist. I prototyped bumping `engineGeneration` in `cancelWarmUpObservation()` (closes the download-phase window Part B widens) but **reverted it** — it only partially closes the race (not the activation sub-window) and would give false confidence. Outcome is benign anyway (Whisper is the correct engine for CJK locales). Belongs in the same Whisper-path-hardening follow-up as the watchdog: gen-guard the save in `activateWhisperEngine` **and** add the watchdog.
- **`onDisappear` doesn't call `stopObservingWarmUp()` / `resetOnboarding()` doesn't clear `engineBusy`** (Greptile outside-diff + second-opinion) — not reachable in production: `OnboardingWindowController.windowWillClose` owns teardown and nils the VM; `resetOnboarding()` is uncalled. Defensive-only; left as-is to keep scope tight.

### Verification
- Full suite green; Swift-6 language-mode build clean; the new test is mutation-verified.

LGTM to merge once CI is green.

PR #573, opened 2026-06-20
Disable transcript Edit in Timed view instead of silently opening raw text

### LINKED ISSUE #570: [Bug Report] I wanted to edit the transcript, but whem i clicked it went only the raw text, i...
## Message

I wanted to edit the transcript, but whem i clicked it went only the raw text, i expected to whem the timed option was checked to edit in the timed version

## Screenshots

Could not upload:
- `Screenshot-2026-06-19-at-23.02.32.png`: `Error: 409 {"message":"is at a880bd470ccb76ad36315ce77c8069ab9c0ff5da but expected 352d846431bb897a1f43ea33614b3632d172c910","documentation_url":"https://docs.github.com/rest/repos/contents#create-or-update-file-contents","status":"409"}`

## System Info

| Field | Value |
|-------|-------|
| App Version | 0.6.23 (20260615125112) |
| Git Commit | `16e3f865f9f6` |
| macOS | 26.5.1 |
| Chip | Apple M5 Max |

---

moona3k:
## What

Fixes #570. With the **Timed** transcript view selected, clicking **Edit** used to silently drop you into the raw-text editor — the timed view you were looking at just vanished, with no explanation.

Now Edit is **disabled while Timed is selected**, with a tooltip pointing you to switch to Text first. No more surprise mode-switch.

```
  Transcript view        Edit button
  ─────────────────────────────────────────────────────
  Text             →     enabled    "Edit the transcript text"
  Timed            →     disabled   "Switch to Text to edit…"
  (no timestamps)  →     enabled    (view is always Text)
```

## Why disable, rather than "fix" the switch

Editing is plain-text-only *by design*. The Timed view is derived on the fly from word timestamps (`TranscriptSegmenter.groupIntoSegments`) and has no editable backing text — only the plain transcript (`cleanTranscript`) is persistable. So the honest move is to gate Edit to the Text view, not quietly swap modes underneath the user.

- `.parakeetAction(.secondary)` is `.bordered`, which dims correctly when disabled — the affordance stays visible and discoverable.
- Editing is reachable from any state: the Text view always falls back to the raw transcript, so no transcript is ever un-editable.
- No save-path change: edits still update `cleanTranscript` and leave word timestamps intact.

## Verification

- `swift build` — clean
- `swift test` — 3922 XCTest (0 failures, 10 skipped) + 16 swift-testing
- This `.disabled(…) + .help(…)` shape is the same one already shipped elsewhere in this view (the Audio menu, disabled with an explanatory tooltip when audio is unavailable), so the macOS tooltip behavior is a known quantity on our macOS 14 target.

### DIFF
--- a/Sources/MacParakeet/Views/Transcription/TranscriptResultView.swift
+++ b/Sources/MacParakeet/Views/Transcription/TranscriptResultView.swift
@@ -1168,12 +1168,20 @@ struct TranscriptResultView: View {
                 .parakeetAction(.primaryProminent)
                 .disabled(transcriptDraft.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty)
             } else {
+                // Editing operates on the plain text transcript only; the Timed
+                // view is derived from word timestamps and has no editable text.
+                // Disable Edit in Timed mode rather than silently dropping the
+                // user into the raw text view when they click it.
                 Button {
                     beginTranscriptEdit()
                 } label: {
                     Label("Edit", systemImage: "pencil")
                 }
                 .parakeetAction(.secondary)
+                .disabled(transcriptDisplayMode != .text)
+                .help(transcriptDisplayMode == .text
+                    ? "Edit the transcript text"
+                    : "Switch to Text to edit. Edits apply to the text transcript; timestamps are preserved.")
             }
         }
     }


moona3k:
Re: the suggestion to wrap the button in a `Group` so the tooltip shows while disabled —

On our macOS 14 target this isn't needed: the same `.disabled(…) + .help(…)` pattern already ships in this very view (the **Audio** menu, which is disabled with a "not available yet" tooltip), and the disabled-state tooltip works there. SwiftUI's `.help()` maps to the AppKit tooltip, which AppKit surfaces regardless of the control's enabled state.

Also, a `Group` wouldn't change anything here: SwiftUI applies a modifier on a `Group` to each child individually, so `Group { Button.disabled(…) }.help(…)` is equivalent to `Button.disabled(…).help(…)` — exactly what we already have. Keeping the simpler form.

PR #574, opened 2026-06-20
Fix live transcript text invisible due to zero-width NSTextView

hungryhenry101:
The TranscriptTextView NSViewRepresentable wraps an NSScrollView containing an NSTextView. Three issues combined to make text invisible even though the view model had correct data (Copy worked, word count increased):

1. No sizeThatFits: the default implementation returned nsView.bounds.size (.zero for a new view), so SwiftUI gave it no frame in the ZStack.

2. NSTextView width never propagated from the scroll view: SwiftUI sized the outer NSScrollView but the document view (NSTextView) stayed at .zero width. With widthTracksTextView=true the text container mirrored that zero width, wrapping every character to its own line — technically present but invisible.

3. Seed a 360pt fallback width in makeNSView so text is visible even before SwiftUI assigns the final frame. Then sync to scrollView.bounds.width in updateNSView and explicitly invalidate textContainer.containerSize to force re-layout at the correct width.

### DIFF
--- a/Sources/MacParakeet/Views/MeetingRecording/TranscriptTextView.swift
+++ b/Sources/MacParakeet/Views/MeetingRecording/TranscriptTextView.swift
@@ -7,6 +7,8 @@ import SwiftUI
 /// Supports drag-selection across the entire transcript with colored speaker headers.
 /// Uses incremental suffix updates so live transcript changes don't rebuild the full document.
 struct TranscriptTextView: NSViewRepresentable {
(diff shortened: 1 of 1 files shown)


moona3k:
Addressed the layout review comments in abf002b2f: switched the NSTextView to native width autoresizing, removed the manual width sync path, and replaced the default 10pt size fallback with a named non-collapsing fallback size.

Local verification on the exact pushed head:
- `swift build`

moona3k:
Thanks @hungryhenry101!

PR #577, opened 2026-06-20
fix(meetings): keep artifact folders visible after audio deletion

moona3k:
## Summary
Meeting artifacts now remain discoverable after retained meeting audio is deleted. Before this, artifact lookup depended on `transcriptions.filePath`, so Delete Audio Only, retention sweeps, and clear-meeting-audio flows could make a saved meeting lose its Finder/CLI artifact path even though `manifest.json`, `transcript.json`, `notes.md`, and prompt results were still useful.

This PR makes the meeting folder identity durable while keeping `filePath` focused on retained playback/export audio:

- adds `transcriptions.meetingArtifactFolderPath` with migration/backfill from existing meeting audio paths
- resolves artifacts from the durable folder path first, with a legacy `filePath` parent fallback
- preserves artifact-folder identity through meeting creation/finalization, materialization, retention, Delete Audio Only, Settings clear-all, and CLI clear-all
- makes full meeting deletion remove the artifact folder even after audio was already detached
- separates artifact UI actions from audio actions in Recent Meetings, Library rows, and saved meeting detail
- keeps CLI meeting JSON/artifact output useful after audio is gone

Out of scope by design: workspace UI redesign, folders/projects, transcript-only mode, cross-meeting Ask, and smoke harness implementation.

## Verification
- `swift test` -> 3950 tests, 10 skipped, 0 failures
- `swift test --filter 'MeetingArtifactStoreTests|TranscriptionDeletionCleanupTests|MeetingAudioRetentionSweeperTests|TranscriptionRepositoryTests|DatabaseManagerTests|MeetingsCommandTests|HistoryCommandTests|TranscriptionLibraryViewModelTests|TranscriptionViewModelTests|SettingsViewModelTests'` -> 521 tests, 0 failures
- `swift test --filter 'TranscriptionDeletionCleanupTests|SettingsViewModelTests/testClearMeetingAudio|HistoryCommandTests/testClearMeetingAudioCommand|MeetingsCommandTests/testMeetingJSONSurfacesArtifactFolderAfterAudioIsGone'` -> 23 tests, 0 failures
- `git diff --check HEAD~2..HEAD`

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT-5


moona3k:
Addressed the three cleanup-loop review comments in `330da3109`:

- bulk meeting audio cleanup now skips unreadable/non-directory entries instead of aborting
- locked meeting folders are logged and skipped during bulk cleanup
- per-folder audio removal failures are logged without stopping cleanup of the remaining folders
- added `testBulkMeetingAudioCleanupSkipsLockedAndFailingFolders`

Verification:
- `swift test --filter 'MeetingArtifactStoreTests|TranscriptionDeletionCleanupTests|MeetingAudioRetentionSweeperTests|TranscriptionRepositoryTests|DatabaseManagerTests|MeetingsCommandTests|HistoryCommandTests|TranscriptionLibraryViewModelTests|TranscriptionViewModelTests|SettingsViewModelTests'`
- `git diff --check`

moona3k:
Follow-up on the cleanup-loop review resolution after the final `0d3829bd` patch:

- per-entry directory metadata still uses `try?`, so unreadable/non-directory entries do not abort the scan
- locked session folders now make bulk cleanup refuse before deleting any audio, rather than skip and continue
- per-folder removal failures now propagate, rather than being swallowed

That is intentional for this caller contract: Settings/CLI clear stored `transcriptions.filePath` only after file cleanup succeeds. Skipping a locked/failing folder and continuing would risk leaving audio on disk while clearing the DB audio pointer for that meeting.

Verification on `0d3829bd`:
- `swift test --filter 'MeetingArtifactStoreTests|TranscriptionDeletionCleanupTests|MeetingAudioRetentionSweeperTests|TranscriptionRepositoryTests|DatabaseManagerTests|MeetingsCommandTests|HistoryCommandTests|TranscriptionLibraryViewModelTests|TranscriptionViewModelTests|SettingsViewModelTests'` (523 tests)

moona3k:
Post-merge fresh-eye follow-up landed directly on `main` in `8d6f8c13c` (`Tighten meeting audio cleanup after PR 577 review`).

Addressed the open Greptile findings:
- Non-standard managed audio: `removeManagedMeetingAudioFiles` now removes top-level regular files with managed meeting-audio extensions, instead of only `meeting.m4a`, `microphone.m4a`, and `system.m4a`; the meeting artifact contract was updated to say this explicitly.
- Stats mismatch: Settings meeting-audio storage stats now use the same managed-audio predicate as cleanup, so non-standard source audio is counted and removable consistently.
- TOCTOU lock window: bulk cleanup keeps the initial all-folder lock preflight, then re-checks each folder lock immediately before destructive removal. That narrows the valid race without changing the caller contract that cleanup must fail rather than silently skip locked folders.
- Partial detach failure: `testDetachMeetingAudioKeepsFilePathWhenFileRemovalFails` now asserts the recoverable invariant that `meetingArtifactFolderPath` is set while `filePath` remains when file removal fails.

Reviewed but intentionally not changed:
- The Gemini suggestions to skip locked/failing folders were not adopted. Callers clear DB `filePath` only after cleanup succeeds, so silently skipping a locked/failing folder could hide retained audio from the DB.

Verification:
- `swift test` — 3,963 XCTest cases + Swift Testing suites passed, 10 skipped, 0 failures.
- `swift test --filter 'HistoryCommandTests|TranscriptionDeletionCleanupTests|SettingsViewModelTests/testRefreshStatsIncludesMeetingAudioStorage|SettingsViewModelTests/testRefreshStatsIncludesManagedNonStandardMeetingAudioStorage|SettingsViewModelTests/testClearMeetingAudioRemovesFilesAndClearsMeetingStoredPaths'` — 37 tests passed.
- `git diff --check` — clean.

PR #598, opened 2026-06-21
Plan: RefinementValidator — guard AI-formatter output, fall back to baseline

moona3k:
## What

Adds an executor-ready plan (`plans/active/2026-06-21-refinement-validator.md`) + status-board row for a **RefinementValidator**: a pure, deterministic guard that checks LLM AI-formatter output against the deterministic baseline transcript and falls back to the baseline when the output looks broken.

**Plan only — no production code in this PR.**

## Why (first principles)

`TranscriptFormatter.format()` is the single chokepoint every AI-formatted lane (dictation, file/URL, meeting) passes through. Today its success path accepts **any non-empty LLM output verbatim** — the only guard anywhere is an empty-string check. There is nothing stopping a small/quantized/local model from dropping half the words, summarizing the transcript away, appending commentary, or looping, and having that persisted as the user's `cleanTranscript`.

This is a **live gap**: file/URL + meeting formatting ships **on by default** (`aiFormatterEnabledForTranscriptions = true`); dictation formatting is opt-in.

The asymmetry that makes this cheap and safe: the deterministic baseline (post-`TextRefinementService`) is *always* shippable, and LLM formatting is optional polish on top. A guard that falls back to that baseline whenever the polish looks broken loses nothing in the bad case and keeps the win in the good case.

## Design (summary)

- **Pure value type** `RefinementValidator` in `TextProcessing/` (no I/O, no LLM call, no throw) with four deterministic checks: `empty`, `lengthRunaway`, `repetitionLoop`, `lowContentOverlap`. Biased toward **accept** so legitimate light/structural formatting is never second-guessed.
- **One insertion point**: the `TranscriptFormatter.format()` success return, per `Lane` (`.dictation` / `.transcription` thresholds). A rejection reuses the existing, already-documented error-fallback return shape (`text: nil` → caller delivers the baseline; provenance not stamped).
- **Default-on kill switch** `AppFeatures.refinementValidationEnabled` (same idiom as `meetingCaptureReliabilityEnabled`).
- **Phasing**: Phase A (pure type + unit tests, zero behavior change, ships alone) → Phase B (wire + flag + formatter tests) → Phase C deferred (rejection telemetry; two-repo allowlist).

## Explicitly out of scope

- **Transforms** (intentional rewrites — summarize/translate/reformat/spoken-transforms). A length/overlap guard would wrongly reject correct Transform output; Transforms never call `TranscriptFormatter`, so they are naturally excluded and must stay that way.

## Verification

Plan doc only; specifies the test matrix (`RefinementValidatorTests` + extended `TranscriptFormatterTests`, regression nets unchanged). No code to build here.

---
## Summary by cubic
Adds an executor-ready plan for a `RefinementValidator` that vets `TranscriptFormatter.format()` output with accept‑biased checks (empty, runaway length, back‑to‑back repetition, stop‑word content overlap) and falls back to the deterministic baseline if broken. Plan only—chokepoint insertion with per‑lane limits, keeps succeeded `LLMRun` on rejection, injectable `isValidationEnabled` defaulting to `AppFeatures.refinementValidationEnabled`, excludes Transforms.

<sup>Written for commit 756f3236dde5a87d48b1624ae578470088dd03ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/598?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


moona3k:
## Review convergence (revision `756f323`)

Two independent model reviewers (Codex/GPT-5.5, Gemini) plus the three PR bots (Greptile, Gemini Code Assist, cubic) converged on the **same four issues**. All are addressed in `756f323`:

1. **Repetition false positives** (Gemini CA, cubic, Greptile-implied) — common phrases like "one of the" recur legitimately in long transcripts. → Changed from "≥N occurrences in a window" to **consecutive / back-to-back n-gram run** detection (the actual LLM-degeneration signature). Dispersed-phrase and scattered speaker-label fixtures added to prove they don't trip.
2. **Kill switch untestable with `static let`** (Greptile P1) — → `TranscriptFormatter` now reads the flag via an injectable `isValidationEnabled` closure (defaulting to `AppFeatures.refinementValidationEnabled`), matching the existing `shouldUseAIFormatter` idiom; the flag-off test injects `{ false }`.
3. **Rejected-run provenance discards metadata** (Greptile P2) — → On rejection we now return the **succeeded** `LLMRun` with `text: nil`, `resolution: nil` (caller still falls back to baseline; provider/model/token data preserved; no fake-failure, no `nil`). Removes the honest-error-type dilemma and its STOP condition.
4. **Tokenizer underspecified** (Greptile P2) — → One shared normalization for both checks: lowercase, split on whitespace, strip *leading/trailing* non-alphanumerics (commas, colons, quotes, markdown `#`/`*`/`-`/`>`/`**`/`_`). Leading/trailing-only means internal punctuation is preserved, so contractions stay intact (`don't` → `don't`, deterministically).

Also: overlap now runs on a **stop-word-excluded content set** so the formatter's intended function-word rephrasing isn't counted as lost content; rejection does **not** post `.macParakeetAIFormatterWarning`; Step 1 **calibrates** limits against realistic accept-fixtures before wiring; Phase C names the queryable-rejection mechanism without reusing `failedFormatterRun`.

Reminder: this PR is **plan-only** — no production code. The above are design corrections to the executor spec.

PR #621, opened 2026-06-28
Add a Settings control for the Cohere compute policy (GPU/ANE)

dudemeister23:
## Summary

The Cohere compute policy — **GPU** (fastest warm latency, but a one-time ~2-min Core ML graph specialization in the background each launch) vs **ANE** (slower warm, no per-launch optimization) — was only settable via the undocumented `cohereComputePolicy` `UserDefaults` key. So every non-technical user is stuck on the ANE default with no way to opt into the faster path.

This adds a **"Cohere Performance"** card to the Engine settings tab (shown only when Cohere is the active engine) with a segmented **Fastest (GPU) / Balanced (ANE)** control.

I measured the gap on an Apple-silicon Mac (Release build), finger-off-`Fn` → text in the field, AI formatter off:

- **GPU:** ~0.6 s per phrase
- **ANE (current default):** ~1.5–2.5 s per phrase

(Background: the latency field report on #602.)

## Behavior

- The card renders **only when Cohere is the active engine**, mirroring the per-engine `engineLanguageCard`.
- The policy is captured once at engine construction (`STTRuntime.ensureCohereEngine` reads `ComputePolicy.current()`), so a flip is **persist-only** — it takes effect on the next Cohere load, not live. (A live reload on GPU would mean a ~2-min foreground "Optimizing…" wait; deferring lets the next launch's existing background warm-up absorb that cost invisibly.)
- To keep the deferred behavior honest rather than silent, when the selection differs from the policy loaded at launch the card shows a **"Not applied yet"** row with a **"Relaunch to apply"** button.
- The relaunch is **guarded — disabled during an active meeting recording**, mirroring the `SparkleUpdateGuard` never-relaunch-mid-recording invariant.
- **Default is unchanged (ANE)**; users who already set `cohereComputePolicy` are unaffected.

## Changes

- `CohereTranscribeEngine.ComputePolicy.save(to:)` persistence helper (paired with the existing `current(defaults:)`).
- `EngineSettingsViewModel`: `cohereComputePolicy` (`didSet` persists + emits `settingChanged` telemetry), a launch-time capture, and the `cohereComputePolicyNeedsRelaunch` computed flag.
- `SettingsView`: the `engineCohereModelCard` ("Cohere Performance") card, the pending row, and the guarded `relaunchToApplyComputePolicy()` self-relaunch.
- `cohere_compute_policy` added to the `settingChanged` telemetry setting names.
- Tests in `EngineSettingsViewModelTests`: default read, persistence round-trip, and needs-relaunch divergence.

## Verification

- `swift build`
- `swift test --filter EngineSettingsViewModelTests` (incl. the new persistence + needs-relaunch tests)
- Full `swift test` — **4,128 tests, 0 failures** (rebased on current `main`, including #613).
- Manual on an Apple-silicon Mac (Release): flipped GPU ↔ ANE and confirmed the pending row appears/clears correctly; the **"Relaunch to apply"** button cleanly relaunches (one instance, no orphans) and the fresh instance loads the chosen policy (`cohere_model_prepare_complete compute=ane` → `variant=ane` dictations after switching to Balanced); the latency difference is real (~0.6 s GPU vs ~1.6 s ANE warm).

---
## Summary by cubic
Add a “Cohere Performance” card in Engine settings to choose Cohere’s compute policy: Fastest (GPU) or Balanced (ANE). The choice is saved, searchable in Settings, and applied on the next Cohere load with a guarded “Relaunch to apply” prompt.

- **New Features**
  - Card shows only when Cohere is the active engine; default stays ANE and respects existing `cohereComputePolicy`.
  - Segmented control persists via `save(to:)`, emits `settingChanged` telemetry (`cohere_compute_policy`), is indexed by Settings search; UI copy clarifies ANE as “Apple Neural Engine (ANE)”; docs/spec updated.
  - “Not applied yet” + “Relaunch to apply” if selection differs from the launch policy; button disabled during an active meeting.
  - Engine reads the policy at construction (`current()`); no live reconfiguration.
  - Tests cover default read, persistence round‑trip, needs‑relaunch flag, and the search index entry.

- **Bug Fixes**
  - “Relaunch to apply” now flushes `UserDefaults` first and quits only after the new instance starts; if relaunch fails, the current app stays running.
  - Added a double‑click guard to prevent spawning multiple replacement instances during relaunch.

<sup>Written for commit 8d8d6461731df8e6b0ae239f35841396c516d898. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


moona3k:
Did a deep-dive review and pushed follow-up polish to get this merge-ready. Thanks @dudemeister23 - the product shape is right.

Validated:
- Cohere compute policy was GUI-unreachable and now has a clear Settings control.
- Persist-now/apply-on-relaunch is correct because the Cohere engine captures compute units at construction.
- The meeting-recording gate matches the app update/relaunch safety model.
- Settings search, telemetry docs, and STT spec now cover the new control.

Follow-up included search coverage, docs/spec updates, a defensive defaults flush before relaunch, and ANE copy/relaunch-overlap polish. Current rebased head is `8d8d646`; full local tests and GitHub CI are green.

moona3k:
Final follow-up on current head `8d8d646`: added the in-flight relaunch guard from review.

The Relaunch to apply action now:
- switches to Relaunching...
- disables while `NSWorkspace.openApplication` is in flight
- resets if launch fails
- prevents duplicate replacement app instances from rapid clicks

With that fix, the PR is rebased on latest main, CI is green, and the review threads are resolved.

PR #639, opened 2026-06-28
docs: architecture deepening opportunities (meeting flow state + STT engine adapters)

moona3k:
## What

A **PROPOSED** `docs/research/` note capturing the two highest-leverage findings from an `/improve-codebase-architecture` review of `main` (`8ab6dddd4`), plus a cross-reference from the STT subsystem README. **Docs-only, zero behavior change.**

The point is to make the *direction* reviewable before any implementation diff exists — and to hand a future agent the **case and the traps**, not just a conclusion.

## Why this is a doc, not a code PR

These are sizeable refactors at load-bearing seams (the STT runtime; the meeting flow). Committing the analysis lets the direction be vetted now; the actual implementation is gated behind interface design (the skill's grilling/design-twice step), tests, and full Greptile + multi-LLM **code** review on the real diff. Greptile is a code reviewer and will have little to say here — the review that matters for a *proposal* is design review, which is already done (see below).

## The two findings

Both share a shape: **the clean pure core already exists** (the meeting flow state machine; the per-engine actor protocols), and the friction is one layer out in how that core is wired/dispatched.

1. **Meeting capture-state transitions are sampled, not pushed.** `MeetingRecordingService.failCapture` flips an internal flag and stops capture but emits nothing; `MeetingRecordingFlowCoordinator` reconstructs the transition by sampling `captureMode` on a 1s poll. No audio is lost in the window (handlers gate on `captureFailed`), so the cost is lost **locality + UI latency + untestability**, not correctness. Direction: push the one missing terminal signal (one-shot continuation, or a lossless transition `AsyncStream`); keep a timer only for continuous metering.

2. **`STTRuntime` answers "what can this engine do?" with ~10 engine switch/if sites + ~15 variant checks.** The adapter pattern already half-exists (`STTTranscribing`, `NativeLiveDictating`); Parakeet TDT is the unwrapped default. Direction: capability-declaring engine adapters; runtime becomes a registry. ADR-016's *decision* is unchanged (internal to the runtime), but its implementation-direction prose needs a touch and capability knowledge straddles the runtime↔scheduler seam.

The note documents alternatives, the **strongest counter-arguments** (the exhaustive `switch` is compiler-enforced safety; a TDT adapter adds a hot-path actor hop; the ANE inference gate must be injected and kept inline), and load-bearing constraints.

## Review provenance

Hardened against **three independent design reviews** (Codex, Gemini, adversarial-document-reviewer) that verified every claim against the code and caught: an overstated "correctness-critical" claim, a false "event-driven vs polling" contrast (dictation polls too), a mislabeled evidence row (`:1785/:1798` are telemetry helpers, not slot routing), and a non-existent "optimistic pause flip." All folded in.

## Scope / non-goals

- No code behavior changes; no ADR added (status is PROPOSED).
- Not a build plan — a detailed plan would be speculative until the seam interface is designed.
- The other six candidates from the review are noted but not written up here.

---
## Summary by cubic
Adds a PROPOSED research note outlining two architecture deepening opportunities: push meeting capture-state transitions instead of polling, and capability-based STT engine adapters with `STTRuntime` as a registry. Cross-links from `Sources/MacParakeetCore/STT/README.md`, fixes Parakeet TDT inline-transcribe line refs in the note, and clarifies the Finding 1→3 numbering; docs-only, no behavior changes.

<sup>Written for commit b84abe897a8b86d75bcf7fd10df330b1cd8f9874. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/639?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Sources/MacParakeetCore/STT/README.md
+++ b/Sources/MacParakeetCore/STT/README.md
@@ -109,6 +109,14 @@ minor historical wart, not a design statement.
   above).
 - `spec/06-stt-engine.md` — narrative spec.
 - Audio buffers feeding the scheduler: `../Audio/`.
+- `docs/research/2026-06-28-architecture-deepening-opportunities.md` — **PROPOSED**
+  (not a decision; constraints not yet validated): collapse `STTRuntime`'s
+  scattered engine-discriminating dispatch into capability-declaring engine
+  adapters (extends the existing `NativeLiveDictating` pattern). ADR-016's
+  *decision* is unchanged, but its implementation-direction prose would need
+  updating and capability knowledge would straddle the runtime↔scheduler seam.
+  Read its constraints section (ANE gate injection, TDT hot-path actor hop,
+  init-serialization guard) before reorganizing engine dispatch.
 
 ## What to know before editing
 
--- a/docs/research/2026-06-28-architecture-deepening-opportunities.md
+++ b/docs/research/2026-06-28-architecture-deepening-opportunities.md
@@ -79,7 +79,7 @@ engines.
 
 Highest leverage of the eight candidates surfaced, both with an in-repo
 precedent that contains the wiring risk, both verified against real code. The
-other six (injectable clock seam; LLM provider adapters; text-processing
+other six (injectable clock seam — Finding 2; LLM provider adapters; text-processing
 orchestration; hotkey matching+validation; preferences write-side; CLI/GUI
 shared policy) are recorded in the session transcript and can be written up if
 prioritized.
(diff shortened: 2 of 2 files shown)


moona3k:
Addressed the review in b84abe897:

- **Inline-transcribe line refs (gemini-code-assist, cubic):** valid — the `:541, :610` I had were unverified and wrong. Corrected to the real pad/URL/preview `manager.transcribe(...)` sites `:504, :554, :687` (each wrapped by `inferenceGate.withExclusiveAccess` at `:503, :553, :686`), verified against the pinned commit. This also makes the "TDT is gated inline" point traceable.
- **Finding 1 → Finding 3 numbering gap (greptile):** valid — added a note under the Finding 3 header explaining the numbers are the original candidate indices (Finding 2 = the injectable clock seam, not written up here), and tagged it in the candidate list so a standalone reader isn't confused.

All other cited line numbers were verified during the pre-PR design-review pass; the doc header notes line numbers are pinned to `8ab6dddd4` and drift over time.

PR #640, opened 2026-06-28
Make Cohere's text-only meeting trade-off visible and recoverable

dudemeister23:
## Summary

Follow-up to the Cohere engine (#602). Cohere — and Parakeet Unified — are word-less (no word timestamps), so meetings transcribed with them are plain text: no speaker labels, no timed segments (the diarization merge is gated on `!words.isEmpty`). That's intentional, but it was **silent**: a user can record a meeting on a word-less engine and end up with an un-diarizable transcript, with no warning and no obvious recovery.

This adds a banner that makes the trade-off **visible and one-click-recoverable** — without changing the engine-per-job behavior #602 chose.

## Change

A **"No speaker labels"** banner above any meeting transcript that has no word timestamps, explaining the trade-off, with a one-click **Retranscribe with Parakeet** that re-runs the retained audio through the existing re-transcribe flow. Falls back to explanatory-only text when the meeting audio is no longer retained.

- Keyed off `Transcription.hasWordTimestamps` (**result-based**), so it's correct for *every* word-less case — Cohere and Parakeet Unified — without a per-engine capability flag.
- Uses `activeTranscription` (live state) and goes straight to the re-transcribe confirmation (no popover), per review.

## Behavior

- No change to transcription or engine routing — visibility plus a shortcut into the existing re-transcribe flow.
- Single file (`TranscriptResultView.swift`, +58).

## Verification

- `swift build`
- Full `swift test` — **4,232 tests, 0 failures**.
- Manual (Release, Apple-silicon): recorded a Cohere meeting and confirmed the "No speaker labels" banner + **Retranscribe with Parakeet** button render above the transcript.

### DIFF
--- a/Sources/MacParakeet/Views/Transcription/TranscriptResultView.swift
+++ b/Sources/MacParakeet/Views/Transcription/TranscriptResultView.swift
@@ -583,7 +583,7 @@ struct TranscriptResultView: View {
         ) { confirmation in
             Button(confirmation.confirmLabel, role: .destructive) {
                 guard confirmation.transcriptionID == transcription.id else { return }
-                onRetranscribe?(transcription, confirmation.speechEngineOverride)
+                onRetranscribe?(activeTranscription, confirmation.speechEngineOverride)
             }
             Button("Cancel", role: .cancel) { }
         } message: { confirmation in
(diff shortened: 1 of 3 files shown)


moona3k:
Thank you @dudemeister23!

PR #683, opened 2026-07-04
fix: polish app-aware formatter readiness

moona3k:
## Summary

App-aware formatter smart defaults are now inspectable before they can run. When the master Smart defaults switch is off, the Settings grid stays readable, category toggles are disabled, and the preview copy makes clear that routing still falls back to the global prompt.

This also aligns the local specs, historical requirement, completed plan, and QA audit with the current release posture: the feature is code-complete, still hidden behind `AppFeatures.aiFormatterProfilesEnabled = false`, and can be flipped later without a schema or migration change.

Related: #117, #412

## Validation

- `git diff --check`
- `swift test --filter LLMSettingsViewModelTests` - 124 tests, 0 failures

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT-5

---
## Summary by cubic
Polished the app-aware formatter UI so built‑in smart‑default prompts stay readable when Smart defaults are off, routing still falls back to the global prompt, and docs/specs align with the current flag‑off posture (`AppFeatures.aiFormatterProfilesEnabled = false`).

- **New Features**
  - Keep prompt previews readable with the master switch off; dim the grid only.
  - Disable per‑category switches until the master switch is on.
  - Use category‑level state for card icon/text colors when the master is off to avoid double‑dimming; improve toggle accessibility to indicate “Enabled, inactive while Smart defaults are off.”
  - Update preview copy to state the prompt will not run while the tier is off and that fallback routing applies.

- **Refactors**
  - Expose category‑level configured state separately from effective routing in the view model, with tests covering the distinction.

<sup>Written for commit 3180beb19e7e37e02a4be8d1f3142d85a0cfc09b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/683?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Sources/MacParakeet/Views/Settings/LLMSettingsView.swift
+++ b/Sources/MacParakeet/Views/Settings/LLMSettingsView.swift
@@ -571,19 +571,17 @@ struct LLMSettingsView: View {
                     smartDefaultCard(categoryDefault)
                 }
             }
-            .disabled(!viewModel.aiFormatterSmartDefaultsEnabled)
             .opacity(viewModel.aiFormatterSmartDefaultsEnabled ? 1 : 0.55)
 
-            // The preview stays readable for individually-disabled categories
-            // (so a prompt can be read before deciding to enable it) — the
-            // footer copy switches to reflect the off state instead of giving
-            // stale "turn this type off" advice.
-            if viewModel.aiFormatterSmartDefaultsEnabled,
-               let selected = selectedSmartDefaultCategory,
+            // The preview stays readable even when smart defaults or an
+            // individual category are off, so a prompt can be inspected before
+            // deciding to let it run.
+            if let selected = selectedSmartDefaultCategory,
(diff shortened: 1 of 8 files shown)


moona3k:
Addressed Gemini's valid styling note in 3180beb19: card foregrounds now use the stored category enabled state while the master switch is off, avoiding double-dimming and preserving category-state legibility. Runtime routing still uses the effective master-and-category policy, and the toggle accessibility value now says when an enabled category is inactive because Smart defaults are off.

PR #899, opened 2026-08-11
Fix AirPods startup and dictation resampling

tsahli:
## What Changed
- Accept an AirPods profile-change notification when a usable buffer proves the tap is attached to the resulting graph, while still rejecting changes after the last usable buffer and System Default changes.
- Configure live dictation resampling for real-time input and retain valid partial output returned when AVAudioConverter consumes a chunk. This prevents 24 kHz AirPods capture from losing roughly one third of speech on the way to 16 kHz.
- Add startup-generation and 24/48 kHz duration regressions, and document both invariants.

## Root Intent
Current main reproduced inputRouteChangedDuringStartup with connected AirPods. After the startup race was fixed, a spoken test stayed on Bluetooth but saved only 5.121 seconds from about 7.85 seconds of input and produced a garbled transcript. The corrected build saved 10.923 seconds from 10.923 seconds of AirPods input and transcribed every spoken word.

## Seed Prompt
Make AirPods dictation reliable without pinning System Default, weakening exact-zero Bluetooth protection, adding capture work to the audio render callback steady state, or changing healthy built-in microphone behavior.

## ADRs Applied
ADR-015 shared-microphone ownership and the existing Audio subsystem routing and render-thread invariants. No public boundary contract changed.

## Author Notes
The focused recorder, startup, tap, and recovery suites pass, as do cold-start and stop-start hardware tests and the end-to-end AirPods dictation. An extreme ten-cycle immediate hardware loop timed out on its fifth reopen; it does not use the production fallback chain and remains a useful separate Bluetooth settle-stress signal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live microphone resampling to preserve partial audio output and prevent dropped speech.
  * Preserved recording duration when capturing from 24 kHz and 48 kHz input devices.
  * Improved startup reliability during Bluetooth and AirPods profile changes.
  * Rejected stale microphone data while allowing valid configuration transitions.

* **Documentation**
  * Updated audio pipeline documentation with resampling and microphone startup behavior.


moona3k (approved):
Thanks, Taylor — this is a careful fix for two real AirPods failures, and the restraint is as valuable as the diagnosis.

On `main`, a normal AirPods profile-change during `start()` looked like a dead graph, so we threw `inputRouteChangedDuringStartup` and walked off the headset the user was speaking into. Once that was unblocked, live 24 kHz → 16 kHz conversion was discarding valid `.inputRanDry` output and dropping a large chunk of the sentence. Your hardware numbers (5.121s saved from ~7.85s spoken; 10.923s in = 10.923s out after) match what the converter math predicts.

The change stays inside the constraints that matter here: System Default stays implicit, exact-zero Bluetooth protection is untouched, the extra generation read is startup-only, and a config change after the last usable buffer still rejects. I replayed the focused audio suites (55/55) and waited on CI; `swift-test` is green.

Walkthrough of the two bugs, the shared-engine architecture, and the fix: https://macparakeet.com/dev/pr/899

Merging. Grateful you chased this through the second bug instead of stopping at “it starts now.”

moona3k:
Walkthrough of the two bugs, the shared-engine architecture, and the fix:

https://macparakeet.com/dev/pr/899

Issue #905, opened 2026-08-13
[Bug Report] Opening a long (~2h) recording from the Library freezes the whole app (synchronous main-thread transcript work on open)


sanjeevpenupala:
## Message

Filed from a source-code review of `main` (Sources at commit `408d1bcd`), reproduced behaviorally on an Apple M3 Pro with ~2 hour recordings. This is a root-cause writeup, not a crash log.

**Symptom:** Opening a long recording (e.g. a 2-hour meeting) from the Library freezes the entire app for several seconds. Short dictations open instantly. In my common case I just want the transcript text and rarely replay audio, but the whole UI locks on open.

**Root cause:** Audio does not appear to be the culprit (it streams lazily via `AVPlayer`; no full-file decode). The freeze looks like a stack of transcript-sized operations run synchronously on the main thread (`@MainActor`) in `TranscriptResultView`'s `onAppear`, before the first frame, each roughly O(word count):

1. `TranscriptionViewModel.loadPersistedContent()` (~`TranscriptionViewModel.swift:1461`, called from `TranscriptResultView` onAppear ~`:248`) re-fetches the full row from SQLite and re-JSON-decodes all `wordTimestamps` / `transcriptSegments` / `diarizationSegments`, even though the Library already handed the view a fully populated `Transcription`. This looks like the heaviest single item.
2. `rebuildSegmentCache()` (~`:247`) groups every word into segments and speaker turns.
3. `currentAIContextText` (~`:252`) builds a second full transcript representation plus a large string join, unconditionally, even when the AI/chat panel is never opened.
4. `mandalaData` (~`:2953`) is a computed property doing an O(word-count) allocation on every `body` re-render (unlike `cachedSegments`, it does not appear to be cached), so it also churns during playback and scroll.

**Amplifier:** reads go through a single serial GRDB `DatabaseQueue`, so the main-thread re-fetch can also queue behind any in-flight background write.

**Why only long recordings:** a 5-minute dictation is ~500-1000 words; a 2-hour meeting is ~20,000. Every pass above is proportionally larger, stacked with no yield on the UI thread, which reads as a whole-app freeze.

The common thread is that this work runs synchronously on the main actor at open time and scales with transcript length. I wanted to surface the mechanism rather than prescribe a fix; happy to share more detail or a repro if that is useful.

Line numbers are approximate to the commit above and may have drifted in the released build.

## System Info

| Field | Value |
|-------|-------|
| App Version | N/A (source-code review of `main`) |
[...]

PR #946, opened 2026-09-03
Fix Transcription view freeze when scrolling down, then scrolling up

### LINKED ISSUE #945: [Bug Report] Scrolling down in Transcription View and then scrolling up causes UI to lock up
When viewing a transcription (even a small one for less than 1 minute of audio) if you scroll down the app gets slightly slow. When you scroll back up the entire UI fully locks up and the app is completely unusable.

This occurs on 0.7.3 and I'm having the same issue when building from the `main` branch.

robd003:
Resolves #945.

## Summary

Opening a transcript from the Library, scrolling down, then scrolling back up could pin the main thread at 100% CPU. The sampled freeze points at a self-feeding SwiftUI layout/update loop involving the `LazyVStack` view cache, realized row heights, selectable-text platform overlays, and hover redispatch.

- Transcripts with up to 400 timed rows now use a plain `VStack`, removing the lazy view cache from the common path.
- Larger transcripts retain `LazyVStack`, preserving the bounded-card protection introduced by #848.
- DEBUG launch overrides (`MACPARAKEET_DEBUG_TRANSCRIPT_LAZY` and `MACPARAKEET_DEBUG_TRANSCRIPT_SELECTION`) remain available for recurrence diagnosis.
- Layout-decision and offscreen scroll smoke tests use an explicit empty environment when asserting default behavior, so developer or CI launch overrides cannot make them nondeterministic.

Robert Palmer's original commit `cfff777f9f233ec454ae79de178fabfa8f108dae` remains intact as the first PR commit. The maintainer corrections are follow-up commits, preserving the contributor's authorship.

## Verification

- `swift test --filter TranscriptBodyLayoutTests` — passed, 4 tests / 0 failures.
- `swift test --filter TranscriptTimestampedLayoutSmokeTests` — immediate rerun passed, 4 tests / 0 failures in 12.382 seconds. The preceding cold run had one threshold-only smoke failure: the 400-row case observed 3 idle layout passes against a ceiling of 2; the other 3 tests passed. No hang or watchdog failure occurred.
- `git diff --check` — passed.
- Hover cannot be simulated in the offscreen host and remains a manual app check.

## Scope and merge order

This PR intentionally remains the standalone small/non-lazy versus large/lazy layout change. Do not add the async-cache work from #905 here; that work should be implemented or rebased after #946 lands so it builds on the finalized layout change.

## Post-Deploy Monitoring & Validation

- Open representative short and long transcripts from Library, scroll to the bottom and back to the top, and confirm the window remains interactive.
- Watch Activity Monitor for sustained main-thread/CPU saturation; healthy behavior is prompt settling after scroll, while sustained 100% CPU or an unresponsive window is the rollback/mitigation trigger.
- Validate both sides of the 400-row boundary during the first release-candidate smoke pass. Owner: release maintainer.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved timed transcript rendering and scrolling, reducing the risk of freezes and excessive CPU usage, especially for longer transcripts.
  - Improved layout stability when viewing and navigating timestamped transcript content.
  - Preserved selectable transcript text with more consistent row-level selection behavior.
  - Small transcripts now use a simpler layout, while larger transcripts use optimized lazy loading.

- **Documentation**
  - Added guidance describing transcript layout behavior and performance safeguards.

### DIFF
--- a/Sources/MacParakeet/Views/Transcription/TranscriptBodyLayout.swift
+++ b/Sources/MacParakeet/Views/Transcription/TranscriptBodyLayout.swift
@@ -16,8 +16,11 @@ enum TranscriptBodyLayout {
     /// The #845 reporter case was 964 rows; typical meetings are far smaller.
     static let nonLazyRowLimit = 400
 
-    static func usesLazyStack(rowCount: Int) -> Bool {
-        if let override = debugOverride(named: "MACPARAKEET_DEBUG_TRANSCRIPT_LAZY") {
+    static func usesLazyStack(
+        rowCount: Int,
+        environment: [String: String] = launchEnvironment
+    ) -> Bool {
+        if let override = debugOverride(named: "MACPARAKEET_DEBUG_TRANSCRIPT_LAZY", environment: environment) {
             return override
         }
         return rowCount > nonLazyRowLimit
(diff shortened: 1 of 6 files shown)


robd003:
@moona3k Any chance you could review this this bug fix and add it to the 0.7.4 release?

I had a major issue with short audio files where when I would scroll down in the transcription and then scroll up the whole app would lock up.

moona3k:
@robd003 Thanks for the fix/PR!

I've been mentally checked out for the past couple months, and I'm now trying to get myself back. MacParakeet 0.7.4 has been long overdue.

PR #955, opened 2026-09-06
Make Discover optional at compile time

alfred-sa:
## Why this is useful

Some local builds should omit the Discover feed entirely. This adds `MACPARAKEET_DISABLE_DISCOVER=1` as a build-time choice while keeping Discover enabled in normal builds. A disabled build excludes its views, services, models, telemetry source case and bundled fallback JSON; the shared Merkaba artwork remains available to the rest of the app.

The distribution scripts isolate build directories by mode and inspect the assembled app and CLI for Discover resources/endpoints. The verifier consumes the complete `strings` output, so a large binary cannot hide a detected endpoint behind a SIGPIPE false negative.

```mermaid
flowchart LR
    F{Disable Discover?} -->|No, default| D[Normal targets and fallback resource]
    F -->|Yes| E[Exclude Discover sources and resources]
    D --> B[Assemble app and CLI]
    E --> B
    B --> V[Verify bundle matches requested mode]
```

## Screenshots

Captured from the corrected integrated app with fictional demonstration data. Navigation may differ from the independent feature branches.

Default build: Discover remains available.

*Default build: Discover remains available.*

Build with MACPARAKEET_DISABLE_DISCOVER=1: Discover is omitted from navigation.

*Build with MACPARAKEET_DISABLE_DISCOVER=1: Discover is omitted from navigation.*

## Scope and review focus

This is an independent extraction from the fork, based on upstream `1159dfca8ae53a15ffcc1562b1efb9220f95cd88`. It contains no prompt, meeting, Markdown or Library layout changes. It is a compile-time distribution option; the existing upstream proposal for a runtime network preference addresses a different use case.

Review both build modes, SwiftPM source/resource exclusion, shared artwork, and packaging checks. Disabled builds must not retain a fallback file, service endpoint or environment override string. The default build must retain its fallback resource.

## Validation

- `bash scripts/dist/test_verify_discover_bundle_mode.sh`: passed on this extracted branch, including app/CLI markers and output larger than a pipe buffer.
- `git diff --check`: passed.
- `MACPARAKEET_DISABLE_DISCOVER=1 swift test --filter 'SpinnerRingViewTests|MainWindowStateTests'`: built this extracted branch and passed all 14 tests.
- The corrected integrated series passed its focused selection: 1,579 tests, one skipped, zero failures. Independent review found no remaining actionable issue in the eight corrections made across the series. This is integrated evidence, not a replacement for both distribution-mode checks above.

Upstream CI currently requires maintainer action (`action_required`), and CodeRabbit skipped review while the PR was a draft. Those checks are not represented as passed.

Local `no-mistakes` and Greptile CLIs are unavailable. The website walkthrough repository referenced by the contributor guide is inaccessible to this account (GitHub returns 404), so the explanation and diagram are included here.

Both corrected Xcode GUI configurations (Discover enabled and `MACPARAKEET_DISABLE_DISCOVER=1`) built and launched successfully for the local visual comparison.

Cubic review follow-up: restored the explicit default-build contract that Discover refresh starts at launch independently of telemetry, with no runtime opt-out. This is a documentation clarification; runtime behavior is unchanged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ journal/
 # Xcode
 .build/
 .build-swift6-no-whisper/
+.build-swift6-no-discover/
 DerivedData/
 /dist/
 *.xcodeproj/xcuserdata/
--- a/Package.swift
+++ b/Package.swift
@@ -5,6 +5,7 @@ import Foundation
 
 let skipWhisperKit = ProcessInfo.processInfo.environment["MACPARAKEET_SKIP_WHISPERKIT"] == "1"
 let enableMLXLocalLLM = ProcessInfo.processInfo.environment["MACPARAKEET_ENABLE_MLX_LOCAL_LLM"] == "1"
+let disableDiscover = ProcessInfo.processInfo.environment["MACPARAKEET_DISABLE_DISCOVER"] == "1"
 
 let packageDependencies: [Package.Dependency] = [
     // GRDB for SQLite (dictation history + transcription records)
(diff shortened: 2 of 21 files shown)


moona3k:
Closing this as superseded, with thanks to @alfred-sa for the investigation.

#965 merged today and makes Discover optional as a runtime preference under Settings > System > Appearance. That covers the use case without needing a second app build. The compile-time approach here was the alternative we weighed, and the runtime toggle's scoping drew on the work done in this PR to isolate the Discover surface.

alfred-sa:
Thanks @moona3k 
I eventually chose compile time because I didn't how much you wanted to keep control on that :)
Config time is better imo. 👍

Issue #662, opened 2026-06-30
Speaker voice profile recognition and naming from historical recordings


yakov0922:
Hi,

I’d like to suggest an optional feature for speaker recognition across historical recordings.

It would be useful if the app could analyze previously recorded audio or meeting recordings, detect recurring speaker voice profiles, and allow the user to assign names to those speakers.

Suggested behavior:

- Detect recurring speaker voice profiles across historical recordings
- Let the user manually name a detected speaker, such as “John”, “Product Manager”, or “Client A”
- Reuse that speaker name in future transcriptions when the same voice appears again
- Provide suggestions when an unnamed voice profile appears multiple times across recordings
- Allow the user to configure a threshold, for example:
  - Suggest naming a speaker after the same voice appears in 3 recordings
  - Suggest naming a speaker after the same voice appears 5 times
- Keep this feature optional, since some users may not need speaker identity tracking

Example:

If the same unnamed voice appears in several past meetings, the app could show a prompt like:

> This speaker appears in 5 recordings. Would you like to name this speaker?

This would make long-term meeting archives much easier to search, review, and organize, especially for users who record recurring meetings with the same people.

This is just a feature suggestion. It would be nice to have, but not critical.

moona3k:
This is a great idea. Let me explore. I can certainly consider upon reviewing whether/how the voice profiling can be done in a accurate, reliable way.

k1n0b0n:
Hey, y'all !

I am currently implementing the plan for this feature, I forked the repo and ran a check about the plan.
I took the liberty to update the plan, mainly to add configuration options for the matching system, to allow calibration on real data specific to the user. Details can be found here :  #993 

I chose to go with "atomic" pull requests to be able to split the review and validation process. It is currently work in progress, and would need a total of 7 PR (+ the one modifying the plan) and I currently did 3. If you think it's preferable, I could merge all that in a single PR.

Also, my skills in swift are limited so I am using Claude code to help me, but I reviewed the code by reading it, and run the tests locally. I intentionnaly added comments in the code to understand it better, but I could also clean that up if necessary.

Looking forward to hear your feedback :)

k1n0b0n:
Hi again !
So I managed to publish all the PRs necessary to implement the feature.
I need to make some human testing now to stabilize the feature.
@moona3k I saw that you merged the first PRs, and that you published some fixes over it, so many thanks 😄 

[...]

PR #994, opened 2026-09-09
feat(diarization): surface speaker embeddings through the diarization adapter

k1n0b0n:
PR 1 of the voiceprints plan (`plans/active/2026-07-03-speaker-voiceprints.md`, as
amended in #993). **No product behaviour changes** — nothing reads the new fields yet.

## What

`DiarizationService` already received a 256-d voice embedding per detected speaker from
FluidAudio and threw it away when building `MacParakeetDiarizationResult`. It now carries
them, alongside per-speaker speech duration — the only gate the matcher will have in v1.

New types in `Sources/MacParakeetCore/Services/Diarization/SpeakerEmbedding.swift`:
`SpeakerEmbedding`, `SpeakerCaptureDomain`, `SpeakerModelIdentity`.

`speechSeconds(forSpeaker:)` is the single conversion point to the unit the duration gates
use. The result carries milliseconds like every other diarization value; dividing at each
call site invites a stray factor of a thousand, which would turn a 3 s gate into 50 minutes
and silently stop every speaker from qualifying.

## Three decisions worth reviewing

**Normalization happens once, at the boundary.** FluidAudio's `speakerDatabase` holds
un-normalized VBx clustering centroids, not unit vectors. Comparing them with a bare dot
product scales every distance by `‖a‖·‖b‖`: for two centroids of norm 0.85 with true
similarity 0.90, the naive distance is 0.350 instead of 0.10 — past any calibrated
threshold, with no error, no log, no crash. The centroid norm also shrinks as a cluster
gets noisier, so the bias is anti-correlated with signal quality. `SpeakerEmbedding.init?`
normalizes and rejects non-finite values and norms below 1e-6.

**Keys are remapped, not copied.** FluidAudio numbers clusters by index; we renumber by
who speaks first. Both use the `S1`/`S2` shape, so carrying the database keys over
untouched would attach one speaker's voice to another's label with nothing to catch it.
`speakerEmbeddings(from:idMapping:identity:)` goes through the same mapping as the
segments, and there is a test that fails if it does not.

**Model identity is two identifiers.** The vector is a clustering centroid, so it moves
when the clustering configuration moves even if the embedding model is untouched — a
change a model-only identifier would miss. `aggregationProfileId` hashes the settings that
can move it. Per-run speaker-count constraints are deliberately excluded: they vary call
to call, and folding them in would make a profile enrolled under a count hint incomparable
with the same voice heard without one.

Speakers whose centroid fails validation — including the all-zero vector FluidAudio emits
when a cluster's responsibility denominator is zero — are dropped from the embedding
dictionary only. They keep their segments, their `SpeakerInfo` and their duration, so
diarization output is unchanged and they are merely unmatchable.

## Tests

24 tests over exact geometry rather than random vectors: a speaker is a basis vector, and
an observation at angle θ has cosine distance exactly `1 − cos θ`, so thresholds read in
degrees.

The one that matters is `testScalingDoesNotChangeDistanceButWouldChangeABareDotProduct`:
it asserts the decision is unchanged **and** that the bare dot product would have differed
by more than 0.15 — so it genuinely fails if normalization is ever removed.

Refs #662

https://claude.ai/code/session_01X2zJL7MPo6Yppd6ned9mAe

## Summary by CodeRabbit

- **New Features**
  - Diarization results now include speaker embeddings for consistent speaker identification.
  - Added per-speaker speech duration totals, with convenient conversion to seconds.
  - Speaker identifiers remain stable across diarization results.
  - Added validation and compatibility checks for speaker embedding data and models.
  - Diarization model identity now reflects relevant configuration settings.

### DIFF
--- a/Sources/MacParakeetCore/Services/Diarization/DiarizationService.swift
+++ b/Sources/MacParakeetCore/Services/Diarization/DiarizationService.swift
@@ -14,6 +14,12 @@ public struct MacParakeetDiarizationResult: Sendable {
     /// exclusive, so these are plain sums.
     public let speechMsBySpeaker: [String: Int]
 
+    /// - Parameters:
+    ///   - speakerEmbeddings: keyed by the same stable ids as `speakers`, and
+    ///     empty when the diarizer produced none. Callers must treat a missing
+    ///     entry as "not matchable", never as an error.
+    ///   - speechMsBySpeaker: total speech per speaker id. Offline segments are
+    ///     exclusive, so these are plain sums.
     public init(
         segments: [SpeakerSegment],
         speakerCount: Int,
--- a/Sources/MacParakeetCore/Services/Diarization/DiarizationService.swift
+++ b/Sources/MacParakeetCore/Services/Diarization/DiarizationService.swift
@@ -6,20 +6,12 @@ public struct MacParakeetDiarizationResult: Sendable {
     public let segments: [SpeakerSegment]
     public let speakerCount: Int
     public let speakers: [SpeakerInfo]
-    /// Voice embedding per detected speaker, keyed by the same stable ids as
-    /// `speakers`. A speaker is absent when its centroid carried no usable
-    /// direction; it keeps its segments and its `SpeakerInfo` either way.
+    /// Keyed by the same stable ids as `speakers`. A speaker is absent when its
+    /// centroid carried no direction; it keeps its segments and label either way.
     public let speakerEmbeddings: [String: SpeakerEmbedding]
-    /// Total speech per speaker id, in milliseconds. Offline segments are
-    /// exclusive, so these are plain sums.
+    /// Offline segments are exclusive, so these are plain sums.
     public let speechMsBySpeaker: [String: Int]
 
-    /// - Parameters:
-    ///   - speakerEmbeddings: keyed by the same stable ids as `speakers`, and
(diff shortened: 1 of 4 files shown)


k1n0b0n:
On the outside-diff finding about `SpeakerModelIdentity`: the VBx half is fixed in the latest push — `vbx.maxIterations` and `vbx.convergenceTolerance` now feed the canonical identity, with a test asserting a change to either yields a different `aggregationProfileId`. They shape the centroid through the posterior responsibilities `computeCentroids` weights by, and folding them in means a FluidAudio upgrade that retunes them is caught without anyone remembering to bump `pipelineRevision` by hand.

The speaker-count constraints are deliberately left out, and the regression test asserting that stays. The constraint is not a configuration constant: `MeetingSpeakerPrior.speakerConstraint` derives it from the calendar attendee count, so it changes from meeting to meeting. Including it would mean a 3-attendee meeting and a 5-attendee one produce different aggregation identities — every profile would then be cross-aggregation against its own samples, `crossAggregationPenalty` would apply permanently, and tau would sit at 0.20 forever rather than the calibrated 0.25. Two enrollments of the same person would also become mutually incomparable.

`aggregationProfileId` is a compatibility marker between representations, not an exact fingerprint of the computation. A per-meeting count hint does not belong in it. The code comment now carries that reasoning with the consequence attached.

Issue #1023, opened 2026-09-13
[Bug Report] Exact speaker count is not honoured: upgrade FluidAudio to 0.15.7 (#891)


k1n0b0n:
## Problem

An exact speaker count requested at retranscription is not held. On a 1:1
meeting retranscribed with *Other speakers = 1*, the system track comes back
with 2 speakers.

Reported from the field in #944: three 1:1 meetings where the single other
participant came back split as *Others* and *Others 1*, worked around with
`--speaker-detection off`.

Upstream cause: FluidInference/FluidAudio #891. The constraint check counts
clusters that win an argmax, while centroid construction keeps every cluster
with `pi > 1e-7`. A cluster that wins no argmax makes the detected count look
already compliant, so no re-clustering happens — and constrained assignment
(pyannote parity: two speakers sharing a segmentation chunk must land on
distinct clusters) then revives the cluster the constraint had excluded. The
upstream PR carries a deterministic counterexample that needs no audio or
models.

The same gate governs `maxSpeakers`, so the `MeetingSpeakerPrior` ceiling
(`max = n + 1`) can be exceeded as well.

Fixed upstream by PR #891, shipped in 0.15.7 (2026-09-10). We pin 0.15.6.

## Fix

- `Package.swift`: move the FluidAudio pin to `exact: "0.15.7"`.
- `DiarizationService.pipelineRevision`: bump to `fluidaudio-0.15.7`. Clustering
  moved, so the aggregation identity must move with it; otherwise existing
  voiceprint exemplars would be scored across a changed clustering.

## Risk

0.15.7 is not a diarization-only release. It also changes ASR: v3 long-form
defaults to the no-mel path (#869), the final streaming window is re-decoded
with a fresh decoder state (#895) and its seam reconciled (#903), custom
vocabulary biasing changed (#898, #899), and NemoTextProcessing became an
opt-out package trait (#892) — the ITN link needs checking. An STT regression
pass is required, not just a diarization comparison (ADR-010).

## Validation

- Before/after on the same audio, same machine: speaker count under an exact
  `--speaker-count N`, determinism across three runs, wall-clock.
- Regression case: a 1:1 meeting retranscribed with *Other speakers = 1*
  returns exactly 1 system speaker.
- Focused tests: `DiarizationServiceTests`, meeting speaker-prior tests, STT
  engine loading tests.
- Full `swift test` once as the final gate.

moona3k:
[...]

Issue #891, opened 2026-08-09
[Feature Request] Extend local-first privacy controls to Discover API calls


adambbutler:
Unlike telemetry, Sparkle, AI providers, and media downloads, there is no preference gating the `macparakeet.com/api/discover.json` network call that happens on every launch.  This appears to be the only gap in the app's otherwise consistent opt-out posture for network surfaces.  A preference there would let it match the rest.

**A note about process, or the lack thereof:** I opened PR #876 for this before reading the README file, because apparently I can dictate faster than I can read.  Filing this retroactively so we can discuss scope and fit, as this would likely be considered a change in behavior.

Happy to rework the PR as appropriate.

moona3k:
Release-readiness scope recorded in #953: Discover and its existing launch-time feed request remain unchanged by explicit maintainer decision for this work. That request is independent of telemetry consent; the docs now state this directly. No Discover opt-out/removal preference is added, and this request remains open rather than being marked implemented.

moona3k:
This landed today through #965, which adapted your #876. Thanks for the original PR.

There is now a **Show Discover in the sidebar** preference under Settings > System > Appearance. When it is off, the app makes no request to the Discover endpoint at launch and cancels any feed work already in flight. Discover stays on by default, and the preference is independent of telemetry consent.

It is on main and will ship in the next app release. 0.7.3 does not have it yet.

Issue #866, opened 2026-07-27
Encourage integration with other tools


mgifford:
I'd be interested in leveraging MacParakeet in a web tool I've built:
https://github.com/mgifford/EchoLocate/blob/main/SPEECH_TO_TEXT_IMPLEMENTATION.md

It would need to be hosted locally, but I do think it could work. 

Right now it is leveraging Chrome, but that's got its limitations.

I don't know if it would be something that would enable this to work with Firefox or other browsers too.

PR #1024, opened 2026-09-13
docs(diarization): plan the FluidAudio 0.15.7 bump for exact speaker counts

k1n0b0n:
Adds the implementation plan for the FluidAudio 0.15.7 bump. **Doc-only: no dependency or source change in this PR**, so the approach can be reviewed before the bump lands.

## Why

A 1:1 meeting retranscribed with *Other speakers = 1* returns 2 system speakers (#1023). The cause is upstream, FluidAudio #891: the speaker-count constraint is checked against the clusters that win an argmax, while centroid construction keeps every cluster with `pi > 1e-7`. A cluster winning no argmax makes the detected count look already compliant, so no re-clustering runs, and constrained assignment then revives the cluster the constraint had excluded. The same gate governs `maxSpeakers`, so `MeetingSpeakerPrior`'s `max = n + 1` ceiling can be exceeded too. Fixed upstream in 0.15.7; we pin 0.15.6.

## What verification already established

This shaped the plan, so it is worth stating up front:

- **No compile break is expected.** The only `ASRConfig` construction here is `ASRConfig(parallelChunkConcurrency: 1)`, and that parameter survives at `v0.15.7`. Nothing references `melChunkContext` (now `melChunkContextOverride`) or `ParakeetEncoderPrecision` (new `int8V2` case) — the unified engine uses a different enum — and there is no exhaustive switch over any changed enum. `OfflineDiarizerConfig` is untouched upstream, so `highAccuracyConfig` compiles unchanged.
- **The real risk is behavioural, not structural.** `CustomVocabularyBoosting.swift` binds the CTC surface that #866/#898/#900 rewrote; signatures survive but `.default` now carries spotter-rescue bounding. v3 long-form flips to the no-mel path (#869), and the final streaming window is re-decoded with a fresh decoder state and its seam reconciled (#895/#903). ADR-010's STT regression pass is therefore required alongside the diarization before/after comparison.
- **`pipelineRevision` re-keys `aggregationProfileId`.** Stored exemplars keep theirs and match at `tau - crossAggregationPenalty` by design; no migration rewrites them and `AppFeatures.voiceProfilesEnabled` is `false`, so there is no user-facing impact today.
- **No boundary contract moves.** `spec/contracts/speaker-voiceprints.md` already documents that the aggregation id is not frozen, so AGENTS.md's "contract change ⇒ contract doc + focused tests" rule is satisfied vacuously rather than skipped.
- **A measurement trap worth knowing.** `RetranscribeCommand.swift:490-505` takes the meeting branch only when `archivedMeetingRecording(for:)` resolves; otherwise it diarizes the mixed track. The decisive test therefore goes through `transcribe --speaker-count 1` on an isolated system-track WAV, with a real archived meeting as secondary confirmation.

## Decisive criterion

The plan's pass gate is a single number: an isolated 1:1 system track with `--speaker-count 1` must return a roster of 1 and one attributed speaker id, against 2 on 0.15.6.

## Notes for review

- `plans/**` is in the CI `paths-ignore` list, so no checks will run on this PR. Expected, not a broken pipeline.
- Status is **TODO — OWNER GATE (audio corpus)** rather than `EXECUTOR-READY`: execution needs a real audio corpus (1:1, three-voice, single-speaker) and human judgement on the WER numbers.
- Local and CI resolve different package manifests (Swift 6.3 picks the new `Package@swift-6.2.swift` with its default-on `NemoTextProcessing` trait; CI on `macos-14`/Swift 6.0 picks the 6.0 manifest), so anything trait-related would break only in CI.

Refs #1023

---
## Summary by cubic
Documents the planned `FluidAudio` 0.15.7 bump for #1023. The current 0.15.6 behavior returns two system speakers for a 1:1 retranscription with *Other speakers = 1*; the bump must return exactly one. This PR is doc-only — no dependency or source change — so the approach can be reviewed before the bump lands.

**Plan**

- Explains upstream #891 and the related `MeetingSpeakerPrior` speaker-ceiling issue.
- Defines an isolated system-track before/after test, plus three-voice and single-speaker corpus checks.
- Pins baseline and candidate binaries explicitly, separates benchmark output, and adds unconstrained runs for B and C.
- Requires compile, diarization, and ADR-010 STT regression checks before updating the dependency.
- Documents `pipelineRevision` re-keying for `aggregationProfileId` without migrating stored exemplars.
- Keeps the work at **TODO — OWNER GATE (audio corpus)**; ASR regressions keep the app pinned to 0.15.6 until a follow-up names the ASR path, defines the override, and lands its acceptance test.

<sup>Written for commit caef822578504857001e1335f8e943205953120f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/1024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Added an implementation plan for upgrading FluidAudio to version 0.15.7.
  - Documented the known diarization issue affecting exact speaker counts and the planned validation approach.
  - Included baseline capture, speech-to-text regression testing, before-and-after diarization checks, acceptance criteria, rollback conditions, and potential CI/toolchain risks.
  - Recorded the work as a high-priority item requiring audio-corpus owner approval.


moona3k:
Superseded by https://github.com/moona3k/macparakeet/pull/1026 (the 0.15.7 pin plus the measured A/B). Fine to close this plan PR whenever you want.

PR #1066, opened 2026-09-16
feat(dictation): optional shortcut to copy instead of paste

### LINKED ISSUE #758: [Feature Request] It would be neat if I could in addition to holding function I could hold maybe s...
## Message

It would be neat if I could in addition to holding function I could hold maybe shift and the dictation would only go to clipboard and would not automatically populate whatever field is active. This is useful because oftentimes I'm on another screen or on another desktop and I'm reading something or looking at something and providing feedback, and then I could switch back to cloud code or whatever I'm using and paste it in place.

## System Info

| Field | Value |
|-------|-------|
| App Version | 0.6.24 (20260621193207) |
| Git Commit | `390e5eeda65b` |
| macOS | 26.5.1 |
| Chip | Apple M2 |

---

moona3k:
## Summary
- Adds an optional **Clipboard-only dictation** shortcut (default unset) so a tap can start/stop dictation and copy the transcript instead of pasting into the focused field.
- Snapshots that destination at recording start so a Settings change mid-utterance cannot start pasting.
- Distinct from *Keep dictation on clipboard*, which still pastes and then leaves a copy behind.

Fixes #758

## Test plan
- [ ] Bind the new shortcut, dictate while another app is focused, and confirm the text is copied and not pasted
- [ ] Use the usual hands-free / push-to-talk shortcut and confirm paste still happens
- [ ] Confirm the new shortcut conflicts with hands-free / push-to-talk (including Command vs ⌘P) and is reserved against Transforms
- [ ] Confirm *Keep dictation on clipboard* still pastes when the extra shortcut is unset

---
## Summary by cubic
Adds an optional **Clipboard-only dictation** shortcut, so a tap can start and stop dictation and copy the transcript instead of pasting it into the focused field. The shortcut is unset by default, so existing hands-free and push-to-talk behavior is unchanged.

- Locks the clipboard-only destination when recording actually starts, so a Settings change or a start requested during processing can't switch it to pasting.
- Treats the shortcut as a separate tap-to-toggle surface and reports conflicts against hands-free, push-to-talk, other transcription hotkeys, and Transform shortcuts.
- Distinct from *Keep dictation on clipboard*: that setting still pastes and leaves a copy behind, while the new shortcut skips paste entirely.
- Drops trailing action keystrokes (such as a press-return snippet) during clipboard-only sessions since nothing is pasted, and logs the skip for diagnosability.

Fixes #758.

<sup>Written for commit 6f9be2061da54276fade72bf88ae68bf12d289cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/1066?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an optional Clipboard-only dictation shortcut.
  - Clipboard-only sessions copy completed transcripts without automatically pasting them.
  - Added settings support, shortcut conflict detection, and shortcut guidance.
  - Clipboard-only dictation preserves the normal recording workflow and supports hands-free or push-to-talk configurations.

- **Documentation**
  - Documented the new dictation mode and telemetry surface.

- **Tests**
  - Added coverage for clipboard-only output, shortcut conflicts, settings persistence, notifications, and telemetry.


moona3k:
Follow-up: the copy-vs-paste choice is now committed when recording actually starts, so a rejected shortcut or pill tap while the last utterance is still transcribing cannot paste a clipboard-only result into the focused field. The shortcut summary in Settings lists the new row, and *Keep dictation on clipboard* no longer says “above.”

PR #1103, opened 2026-09-19
fix(audio): stop Instant Dictation mic re-acquisition loop on macOS 27 (#1102)

### LINKED ISSUE #1102: [Bug Report] # MacParakeet 0.8.7 — microphone is re-acquired ~1.3 times per second while the ...
## Message

# MacParakeet 0.8.7 — microphone is re-acquired ~1.3 times per second while the app is idle

## Environment
- macOS 27.0 (build 26A428), Apple Silicon. macOS 27 installed 2026-09-03.
- MacParakeet 0.8.7 (CFBundleVersion 20260918190255), /Applications/MacParakeet.app
- Input device: BuiltInMicrophoneDevice (selectedMicrophoneDeviceUID)
- Relevant prefs: instantDictationEnabled = 1, showMenuBarIcon = 1, speechRecognitionEngine = whisper, menuBarOnlyMode = 0

## Symptom
An orange microphone indicator in the menu bar blinks roughly once per second for as long as the app is open, including when it is completely idle (Transcribe tab, no dictation, no meeting recording).

## What the logs show (log show, predicate: process == "MacParakeet")
An AVAudioEngine teardown/rebuild cycle repeating at ~1.3 Hz (period ~750-800 ms). Exact 1:1 correlation over a 10-minute session:

    328  AVAudioEngine.mm:1282  "start, was running 0"
    327  AVAudioEngine.mm:1552  "iounit configuration changed > posting notification"

Per-cycle sequence, with a fresh AVAudioEngine instance each time (addresses are reused, so the engines are being freed — this is churn, not a leak):

    Engine@0xXXXX: start, was running 0
    Engine@0xXXXX: iounit configuration changed > posting notification
    Engine@0xXXXX: stop, was running 1
    Engine@0xYYYY: start, was running 0

Each cycle re-acquires the microphone at the HAL level, which is what toggles the privacy indicator:

    setPlayState Started  Input {BuiltInMicrophoneDevice, 0x..}
    setPlayState Stopped  Input {BuiltInMicrophoneDevice, 0x..}
    -> 327 starts / 646 stops in 10 minutes

Each cycle also throws, and issues a TCC request:

    1312  (CoreAudio) throwing -10877      # kAudioUnitErr_InvalidElement (4x per cycle)
     487  (TCC) TCCAccessRequest() IPC

Device churn inside one cycle, on a single AUHAL instance (0x7cbfced540), driven by the app ("caller requesting device change"):

    AUHAL.cpp:2299  SetProperty: caller requesting device change from 75 to 130
    AUHAL.cpp:2299  SetProperty: caller requesting device change from 130 to 70

      device 75  = BuiltInSpeakerDevice              (Input:No  | Output:Yes)
      device 130 = CADefaultDeviceAggregate-<pid>-0  (Input:No  | Output:No)
      device 70  = BuiltInMicrophoneDevice           (Input:Yes | Output:No)

The four -10877 throws occur while the AUHAL is still on device 75/130, i.e. on devices reporting 0 input streams, before device 70 is selected. Timestamped extract of one cycle:

    22:14:31.224  Engine@0x7cbb5411c0: stop, was running 1
    22:14:31.238  setPlayState Stopped Input {BuiltInMicrophoneDevice, 0x1b}
    22:14:31.241  SetProperty: caller requesting device change from 75 to 130
    22:14:31.241  IsDeviceUsable: Device ID: 130 (Input:No | Output:No): true
    22:14:31.274  (CoreAudio) throwing -10877   x2
    22:14:31.275  (CoreAudio) throwing -10877   x2
    22:14:31.294  SetProperty: caller requesting device change from 130 to 70
    22:14:31.295  IsDeviceUsable: Device ID: 70 (Input:Yes | Output:No): true
    ... engine starts again, mic re-acquired ...

## Two candidate mechanisms (not resolved from the outside)
1. A configuration-change feedback loop: the handler for AVAudioEngineConfigurationChange unconditionally recreates and restarts the engine, and starting the rebuilt engine emits another configuration change.
2. A watchdog/retry: the ~800 ms gap between "iounit configuration changed" and the following "stop" is long for a direct notification handler, so a timer that tears the engine down after the input tap has yielded no/invalid data (cf. the four -10877 per cycle) fits the timing at least as well.

System input processing does attach to the client, but only once at launch (22:13:59), not per cycle — so it is a plausible source of the first configuration change, not of the repetition:

    System_Input_Processing_Notification_Handler::is_vi_available(): Voice isolation
      DSP is available for client with bundle id com.macparakeet.MacParakeet
    (CoreAudio) HALSystem.cpp:135  Enabling HAL Voice Isolation support
    AVAUVoiceIOChatFlavor vpio_preferenceChangedListener: new value (2) for key
      AUVoiceIOClients/com-macparakeet-MacParakeet/AUVoiceIOChatFlavor

The recurring per-cycle evidence is all app-side ("caller requesting device change").

## CONFIRMED TRIGGER: Instant Dictation

Turning Instant Dictation off stops the loop. In a session launched at 22:36:06 with the setting still on, the loop ran at the usual ~1.3 Hz. The setting was toggled off at 22:36:53 (AppKit sendAction:), and the loop ended a few seconds later:

    22:36:58.921  setPlayState Started Input {BuiltInMicrophoneDevice}
    22:36:58.922  Sent updated IOState to server: [1, 0]
    ... no further engine start / stop / -10877 for the remaining 80 s of observation

Counts before vs after the toggle, same app session:

    22:36:06 - 22:36:58 (52 s, setting ON) :  ~68 engine starts, ~68 mic acquisitions
    22:36:59 - 22:38:18 (80 s, setting OFF):    0 engine starts,   0 mic acquisitions

So the restart loop lives in the Instant Dictation "armed" path, not in the general audio setup. Note the end state is IOState [1, 0] — input running and held open continuously. The microphone is still in use, it just stops being re-acquired, so the indicator goes from blinking to steady rather than turning off.

### Cold launch with the setting already off

The app was quit (instantDictationEnabled flushed to 0) and relaunched at 22:40:45, then left idle on the Transcribe tab for 98 seconds:

    engine starts (AVAudioEngine.mm:1282)              : 0
    microphone acquisitions (BuiltInMicrophoneDevice)  : 0   (string does not appear at all)
    iounit configuration changed                       : 1   (once, at launch)
    throwing -10877                                    : 4   (once, at launch)

So with Instant Dictation off from the start, the microphone is never opened — cleaner than toggling mid-session. System input processing still attaches once at launch with no microphone IO following it, which further indicates the OS-side attach is not what drives the loop.

Summary of the three measured states, all idle on the Transcribe tab:

    Instant Dictation ON                  : ~1.3 mic acquisitions/second, indefinitely
    toggled OFF mid-session               : loop stops within ~6 s, mic left open (steady)
    OFF before launch (cold start)        : microphone never opened

## Suggested investigation
- Guard the AVAudioEngineConfigurationChange handler against re-entrancy, and compare the new hardware format against the current one before rebuilding — rebuild only on an actual change.
- Keep one long-lived AVAudioEngine with the input tap attached instead of allocating a new engine per cycle.
- The -10877 (kAudioUnitErr_InvalidElement) suggests the input node is configured against an element/device with 0 input streams (device 75 / 130 above) before the microphone device is selected; ordering device selection before format/tap setup may remove both the throws and the restart trigger.
- Holding the microphone open while armed appears intended (settled state is IOState [1, 0]) — so the defect is the re-acquisition loop, not the open mic itself.

## Ruled out
- Not a stale build: 0.8.7 was built 2026-09-18, after macOS 27.0 shipped, and is the newest entry in https://macparakeet.com/appcast.xml (0.8.4-0.8.7 all shipped within 48 hours).
- Not new in 0.8.7: the same loop appears in the 2026-09-18 session logs (329 engine starts) under the previously installed build.
- macOS 27 specificity unresolved: unified-log retention on this machine only reaches 2026-09-18, and macOS 27 was installed 2026-09-03, so no pre-27 session is available for comparison.

## System Info

| Field | Value |
|-------|-------|
| App Version | 0.8.7 (20260918190255) |
| Git Commit | `f8f62dd28876` |
| macOS | 27.0.0 |
| Chip | Apple M5 Pro |

---

moona3k:
## What & why

Fixes the microphone re-acquisition loop reported in #1102: with **Instant Dictation** enabled on **macOS 27**, the shared mic engine is torn down and rebuilt ~1.3×/second while the app is idle, blinking the mic privacy indicator continuously, throwing `kAudioUnitErr_InvalidElement` four times per cycle, and clearing the pre-roll ring each cycle.

**Root cause — a NotificationCenter feedback loop.** While Instant Dictation is armed, `AudioRecorder` holds a passive *warm* subscriber that keeps the shared mic engine running while idle. On macOS 27 that running engine self-emits an `AVAudioEngineConfigurationChange` as it selects its input device. The configuration-change observer's running/non-`prepared` branch re-broadcast *every* such change as `macParakeetMicrophoneSelectionDidChange`; the app answers by tearing down and rebuilding the warm engine, whose fresh start self-emits another change → loop. The 0.5 s warm-refresh debounce only paces it (~0.5 s + ~0.3 s rebuild ≈ 1.3 Hz).

The `prepared` branch of the same observer already absorbs self-emitted changes; the running/committed branch had no equivalent guard. Latent since #862 (2026-07-22), activated by macOS 27's start-time device churn.

Full root-cause write-up: `docs/audits/2026-09-19-issue-1102-instant-dictation-mic-reacquire-loop.md` (in this PR).

## The fix

Give the observer's running branch a compound guard that mirrors the `prepared` branch. When a **still-running** engine receives a self-emitted change whose **input format and resolved route are unchanged** on a **positively non-Bluetooth** input, absorb it (log `configuration_change_ignored reason=unchanged_running_route`) instead of posting + rebuilding.

| Situation | Behavior |
| --- | --- |
| Same route + format, non-Bluetooth, still running | **Absorbed** (loop broken; pre-roll preserved) |
| Real format change (#796) | Posts |
| Real route / default-device change (#796) | Posts |
| Bluetooth / unresolved route (#862/#481 transport flips) | Posts |
| Engine stopped itself (recovery) | Posts + recovers (unchanged) |

- Committed format is read from the committing tap's first buffer (no extra HAL query → no start-path latency); route is the already-resolved snapshot; Bluetooth-safety via `preparedAttemptIsSafe`. Cleared on every engine teardown/replace.
- Adds an `engineRunningProbe` test seam (default = real `AVAudioEngine.isRunning`) so the running branch is unit-testable without hardware.

**Rejected alternative:** a recorder-level compare-before-rebuild (the review's first choice) is unsafe — a warm refresh also legitimately rebuilds to drop VPIO after a meeting leaves, unrelated to input format, and the recorder lacks the VPIO state to distinguish. The observer is the correct layer. Details in the audit doc.

## Tests

- `swift test` full suite: **6961 tests, 0 failures** (24 skipped).
- New in `MicrophoneEnginePlatformConfigChangeRecoveryTests`: unchanged route → **no** post/rebuild; changed route → posts; changed format → posts; Bluetooth route → posts.
- Reviewed with Fable 5.1 (root-cause + implementation diff); findings folded into the audit doc.

## ⚠️ Merge gate (one external confirmation)

This guard only fires if macOS 27's self-emitted change arrives while `AVAudioEngine.isRunning == true`. Apple's contract is that the engine stops itself *before* posting; the report's per-cycle log sequence (`configuration changed` → later `stop, was running 1`) is consistent with it still running, but this should be confirmed from the reporter's `shared_mic_engine_configuration_changed` log line's `engine_is_running=` field before merge.

**If `engine_is_running=false`:** the same committed-state comparison must instead gate the post in the stopped-engine branch (dropping the `engineIsRunning` requirement). The device-selection-ordering cleanup for the `-10877` churn is a separate follow-up that needs macOS 27 hardware.

Closes #1102. Confirmation of the `engine_is_running=` value from the reporter's `dictation-audio.log` is requested on the issue; if it comes back `false`, the stopped-engine branch gets the same committed-state gate in a follow-up.

---
## Summary by cubic
Fixes #1102 by stopping the Instant Dictation microphone teardown/re-acquisition loop on macOS 27. Previously, the running engine rebroadcast its own unchanged configuration changes, rebuilding the warm engine and clearing pre-roll; now those changes are absorbed while real route, format, and Bluetooth changes still follow the existing handling.

**Bug Fixes**
- Snapshots the first usable input format and committed route, with Bluetooth safety evaluated at notification time.
- Adds a deterministic engine-running probe and regression tests for unchanged, changed, and Bluetooth routes using the live input format.
- Documents the feedback loop, start-ordering behavior, and why a recorder-level fix was rejected.

**Merge gate**
- Confirm the reporter logs show `engine_is_running=true`; if not, apply the same committed-state guard to the stopped-engine branch before merging.

<sup>Written for commit d502c57212b2abd4dc6cd4a541ee5909ec127f08. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/1103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved microphone stability during Instant Dictation on macOS.
  * Prevented unnecessary microphone engine restarts when audio configuration remains unchanged.
  * Preserved recovery when input formats, routes, or Bluetooth audio devices change.
  * Reduced microphone reacquisition interruptions and related recording disruptions.

* **Documentation**
  * Added a root-cause review documenting the microphone reacquire issue, affected conditions, and resolution.


moona3k:
## Fresh-eye review (at `5b2ef55`, follow-ups pushed as `d502c57`)

**Root cause: confirmed.** Read the observer end to end; the `prepared`-branch absorb vs. unguarded running-branch post asymmetry is real, the cadence math holds, and the fix absorbs strictly less than before (same route + same format + non-BT + engine still running). Fail-open everywhere it matters; committed state is recorded at all three commit sites and cleared on every teardown/replace. Rejecting the recorder-level fix is right (VPIO-drop rebuilds must survive).

**Verified locally**
- Focused suites (`ConfigChangeRecovery`, `RealPlatform`, `SharedMicrophoneStream`, `AudioRecorderFormatChange`): 113 tests, 0 failures.
- Mutation-checked the negative tests: with the route left unchanged, the changed-route test now fails (it previously could only fail on 48 kHz/1ch hardware); with the BT flag cleared, the Bluetooth test fails. So they test the axis they claim.

**Changes in `d502c57`**
- Route/Bluetooth tests now commit the engine's live input format via one helper, so they cannot pass on a format mismatch.
- Mid-startup variant (raised in the audit's Fable pass): not reachable. `configureAndStart` runs in `queue.sync`, the observer body is `queue.async` on the same serial queue, and a failed start swaps `audioEngine`, so a change emitted mid-start is handled after commit (guard applies) or dropped by `engineBox.wraps`. Documented in a code comment and the audit; no third guard added.
- Audit doc: fixed anchors (CodeRabbit), dropped uncommitted `/tmp` transcript paths.

**Still open before merge (external)**
1. `engine_is_running=` confirmation from the reporter's `dictation-audio.log`. The guard requires the engine to still be running when the macOS 27 change fires. The reporter's `stop, was running 1` after the change is consistent with that, but is not conclusive. If it is `false`, this PR is harmless but does not close #1102.
2. Green `swift-test` on `d502c57`.

PR #1104, opened 2026-09-20
Add experimental Voice Control: Jev judges landings, MacParakeet acts

moona3k:
## Summary

This PR adds a DEBUG-only Voice Control experiment to MacParakeet: ordinary speech or a typed inbox command drives the app already in front of you through native macOS Accessibility. Jev is the judge of competing landings. MacParakeet owns observation, legality, execution, and verification.

It is not a stable-release claim. Live Google Flights results and the integrated microphone are still unproven. Enable with `--enable-voice-control` in a DEBUG build.

---

## What Jev is

Jev (TypeSafe, pinned here at `jev-1.13.0`) is **System One judgment**, not an agent. It does not browse. It does not write AppleScript, JavaScript, or selectors. It does not observe the Mac.

It **can** choose among observed landings given the labels we send it. That is a verification-shaped Choice: after the host acts, which of these named states holds? It does not issue receipts. `finished` from Jev is a judgment over that text, not proof the user’s goal succeeded. Confidence is concentration of the distribution, not P(the search completed).

One request is `state` plus named `questions`. The questions share that state, run **independently**, and cannot read each other’s answers. The primitive that matters for us is **Choice**: a closed map of options, each with criteria, returning a distribution and a confidence.

Jev is strong at picking among named options given evidence. It is weak at arithmetic, dates, counting, generation, huge irrelevant trees, and **simulating a multi-step transition function**. That last weakness is the whole product question.

A practitioner put it cleanly: *predict outcomes, not steps toward an outcome.* Tetris makes the granularity obvious:

| Ask Jev | What you are really asking |
| --- | --- |
| “Play Tetris” | Plan and execute. This fails. |
| “Which button to press” | The next micro-step. Better, still a simulation. |
| “Where should this piece land” | An observed landing. The host compiles left/rotate/drop. |
| “Should you win or lose” | Unbounded completion. If we could compile the goal to a check, we would run the check. Jev picking `finished` over truncated labels is not that check. |

The unsolved “tree where one outcome changes the next” is not a Jev feature. Execute one compiled landing, re-observe, offer a new independent Choice. The graph lives in the host.

Canonical write-up: [decision architecture](docs/research/2026-09-19-jev-voice-control/jev-decision-architecture.md) and [native Accessibility direction](docs/research/2026-09-19-jev-voice-control/native-accessibility-direction.md).

What we refused to build: a seven-state universal Mac graph, Score-ranking every widget, a generative worker in the click loop, CDP, or treating Jev `finished` as a receipt.

---

[...]

PR #1108, opened 2026-09-21
Voice Control: perception as a pure, tested walk; off-screen controls by name

moona3k:
## Design

Observation should be a pure function over the accessibility tree, so its pruning rules can be tested against a fake tree and its cost measured per node. `NativeVoiceControlAdapter.observe()` was a 160-line method interleaving AX IPC with policy: untestable without a live app, ~15 attribute reads per node, `CGGetActiveDisplayList` per node. It is now:

```
acquire process → AXTreeWalk.run(roots, LiveAXTreeSource, display, window, focused, caps)
              → candidate pass (values, settable, selection, fingerprint — kept nodes only)
              → apps + destinations + undo as before
```

Stacked on #1104; independent of #1107. Plan: `plans/active/2026-09-20-voice-control-observability-perception-decision.md` (on #1107).

## Contracts

**`AXTreeSource`** — `children(of:)` and `facts(of:) -> AXWalkFacts { role, label, frame?, hidden, pressable, menuOpen, text? }`. Production wraps `AXUIElement`; tests use a dictionary.

**`AXTreeWalk.run(roots:source:display:window:focused:caps:now:) -> AXWalkResult { candidates, offscreen, staticText, visited, complete }`**, depth-first, with these rules, each covered by a fake-tree test:

| Rule |
|---|
| A hidden node ends its subtree |
| A closed menu bar item's children are never walked |
| A real frame wholly off the display ends visibility for its subtree; labelled pressables inside are still collected as off-screen controls until the cap |
| A node under 4 pt on either side is a sliver, never visible |
| A bare child borrows its parent control's label and is emitted once; rows, cells and buttons take a shallow static-text or image name |
| A nameless `AXGroup` is never a candidate, even when pressable |
| The same role, label/text and frame is one control however many objects the bridge hands over |
| Node, time, children and depth caps report `complete == false` |
| Editable, focused, scroll-area and pressable-static-text nodes are candidates without a label; other unlabelled press-only nodes are dropped |

**`VoiceControlTarget.isOffscreen`** — labelled pressables the app exposes but does not show (Notes rows parked far below, Chromium links clamped to slivers, a hidden Dock). `AXPress` does not need visibility. Reachable by exact spoken name only: `VoiceControlLegality.offeredTargets` excludes them from every Jev request; `execute` skips the visibility guard for them (fingerprint, foreground and window checks stay); a press is received as `unknown` unless transition evidence changes. Deduplicated against visible labels so one name never splits. Custom decoder keeps older `latest.json` fixtures loading.

Display bounds and the window frame are read once per observation. Default node cap 600 → 1,200 (hard 2,000); 2 s budget unchanged; the #1107 timing line is how the effect gets measured.

## Test evidence

```
swift test --filter 'AXTreeWalk|VoiceControl|DictationFlowCoordinator|TransformRunSerializer'
214 tests, 0 failures
```

New: `AXTreeWalkTests` (20). Not yet done: a live observe on Chrome/Finder/Notes with the new caps; numbers will be posted here once #1104 and #1107 are in.

## Behaviour changes

Unlabelled icon buttons are no longer targets (they were `label: ""`; nothing could address them).

## Summary by CodeRabbit

- **New Features**
  - Improved Voice Control discovery with bounded accessibility-tree scanning and clearer candidate filtering.
  - Supports exact-name pressing of labelled controls that are currently off-screen.
  - Adds off-screen status to voice-control targets while preserving compatibility with existing data.
  - Improves label recovery and handling of focused, editable, pressable, and web controls.

- **Bug Fixes**
  - Prevents hidden, duplicate, closed-menu, and invalid-size elements from appearing as targets.
  - Excludes off-screen controls from suggested targets and remote requests.

- **Documentation**
  - Documented accessibility scanning, off-screen control behavior, and related contracts.


moona3k:
## Live observe numbers (dev build with this walk + #1107 timing line)

Frontmost app, `dryRun` submit via the inbox, single observation each, screen text (#1110) enabled and included in the observe phase:

| App | controls | complete | observe |
|---|---|---|---|
| Google Chrome (Gmail search) | 158 (26 buttons, 27 links, 53 static, 9 groups, 4 popups…) | yes | 844 ms |
| Notes | 87 (38 links, 10 buttons, 2 menus…) | yes | 620 ms |
| Finder (list view) | 121 (19 text fields, outline, 8 menus…) | **no** | 3,078 ms |

Two things this surfaced and fixed on the branch (`ff…` "never de-duplicate nameless containers"): the first Chrome observe returned **5** controls because the dedupe key applied to nameless containers and web layouts nest same-sized nameless groups — the second one's subtree was pruned. Regression test added; the 158 above is after the fix.

Finder is over budget because the OCR pass (#1110) runs inside the same observe and Finder's list view is dense; AX alone was under 2 s. Follow-up on #1110: run OCR concurrently with the walk, or give it its own budget.

PR #1110, opened 2026-09-21
Voice Control: on-device screen text as a second observation source

moona3k:
## Design

Accessibility is the primary, authoritative way Voice Control sees the Mac. It is not the only way the user sees it: prices, dates, status lines, list rows, terminal output and canvas text are often absent from the tree or exposed as unlabeled static text. This PR adds **on-device Vision OCR of the frontmost window as an optional second source**, opt-in per user because it needs Screen Recording.

Two outputs from one pass:

1. **Targets.** Recognised lines that no Accessibility control explains become `role: "text"` press targets with a private pixel centre. Accessibility handles are always preferred; the pixel click is the fallback only when no handle exists, and its receipt is `unknown` unless transition evidence changes. Offered only on a plain surface, never inside a picker.
2. **State.** The same lines join the snapshot `summary` under `Screen text:` (reading order, deduplicated against AX static text, 4,000-char cap unchanged), so Jev and local matchers see what the user sees.

Base: `main` (#1108 merged). ADR-033 amendment included. Plan: `plans/active/2026-09-20-voice-control-observability-perception-decision.md`.

## Contracts

| Contract | Where |
|---|---|
| `ScreenTextBlock { text, confidence, frame }` in screen points | `ScreenTextSource.swift` |
| `ScreenTextReading.read(window:) -> [ScreenTextBlock]`; `[]` when Screen Recording is denied | `ScreenTextSource.swift` |
| `VisionScreenTextReader` actor: capture → Vision `.accurate` → point mapping; unchanged-frame skip via a 1/8-scale grayscale thumbnail; nothing persisted | `ScreenTextSource.swift` |
| Reader runs on its own actor concurrently with the walk, awaited for at most 1.2 s (a slow pass keeps running and is cached for the next observation) | `NativeVoiceControlAdapter.observe` |
| Frames of this process's own windows are excluded (the panel's instruction text is never screen text); glyph-only blocks (`f`, `#`, `•`) need two alphanumerics to count | adapter, `ScreenTextMerge.isAddressable` |
| `ScreenTextMerge.readingOrder / mergeLines / matches / unexplained` (secure frames and secure words dropped) | `ScreenTextSource.swift` |
| `NativeVoiceControlAdapter(screenText:)`; `textTargets(from:controls:excludedFrames:existingText:)`; `lastObservationStats` | `NativeVoiceControlAdapter.swift` |
| Text-target press: marked click inside `authority.perform`, then the existing transition loop | `NativeVoiceControlAdapter.execute` |
| `VoiceControlSituation.classify` ignores text targets; `offeredTargets` excludes them in pickers | `VoiceControlMachine.swift` |
| `AppFeatures.voiceControlScreenTextDefaultsKey`; setup toggle; coordinator requests Screen Recording once | app, view model, panel |
| ADR-033 observation amendment; contract; Core README; tools | docs |

## Privacy and permissions

Pixels never leave the Mac. No image is written to disk, even for local debugging; the only retained pixels are the thumbnail bytes inside the actor. Screen text is local-log only and never enters shareable diagnostics. Lines inside secure fields' frames and lines containing the secure-word list are dropped before they become targets or state. Denied permission degrades silently to Accessibility-only observation. Default-on is explicitly not decided; that waits for replay-corpus measurements (#1107) of latency and clarify rate with and without screen text.

## Test evidence

```
swift test --filter 'ScreenText|AXTreeWalk|VoiceControl'
179 tests, 0 failures (after merging #1108 and the picker fixture fix)
```

New: `ScreenTextSourceTests` (7). Live-verified on the dev build with Screen Recording granted: Chrome (Gmail) 14 unexplained blocks after AX explained the rest, Notes 12, Finder 80; details in the comment below. The denied-permission path is reasoned from `CGPreflightScreenCaptureAccess`, not exercised live.

## Known trade-offs

`CGWindowListCreateImage` is deprecated on macOS 14 (ScreenCaptureKit is the successor); it compiles with a warning and works. `.optionOnScreenOnly` reads whatever is visible inside the window rect, overlapping windows included (our own windows are excluded). Tile-level re-OCR is a later optimisation. The adapter is constructed once at launch, so toggling requires End/Start of Voice Control.

---
## Summary by cubic
Adds on-device Vision OCR of the frontmost window as an optional, per-user second observation source for Voice Control, so commands can reach prices, dates, and result rows the Accessibility tree never labels. Accessibility stays the primary source; a pixel click is the fallback only when no AX handle exists.

**Key behaviors**
- Recognised lines that no AX control explains become `role: "text"` press targets with a private pixel centre; receipts are `unknown` unless transition evidence changes. Targets are offered only on a plain surface, never in a picker — and picker detection now ignores text targets, so a focused row whose label merely contains a comma no longer counts as picker evidence without overlay chrome.
- The same lines join the snapshot `summary` under `Screen text:`, deduplicated against AX static text, within the existing 4,000-character cap.
- Opt-in via `voiceControl.screenText.v1`; toggling on requests Screen Recording once, and denial degrades silently to Accessibility-only observation. The toggle needs a Voice Control End/Start to apply.
- Secure-frame lines, secure-word lines, single-glyph or symbol-run blocks, and this app's own windows are excluded from capture, since it reads every pixel inside the window rect.
- OCR runs on its own actor under a 1.2 s budget while the Accessibility walk proceeds; a slow pass never delays observation, and its result is cached for the next unchanged frame.
- Pixels never leave the Mac and nothing is persisted. The capture uses the deprecated `CGWindowListCreateImage` (compiles with a warning; ScreenCaptureKit is a later swap).
- Live measurements (dev build): Chrome left 14 unexplained screen-text blocks after Accessibility, Notes 12, Finder 80.

<sup>Written for commit 957d51129367c32fdb954649d54dbbcad9ab15e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/moona3k/macparakeet/pull/1110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Voice Control can optionally read visible on-screen text using on-device Vision OCR.
  - OCR-detected text can appear in summaries and serve as pressable targets when Accessibility controls are unavailable.
  - A new setting lets users enable or disable this capability per Mac.
- **Privacy & Permissions**
  - Screen Recording permission is required; content remains on the Mac and is not persisted or transmitted.
  - Secure fields and sensitive text are excluded.
- **Bug Fixes**
  - Screen-text targets are excluded from picker interactions.
- **Tests**
  - Added coverage for OCR ordering, matching, filtering, and target behavior.


moona3k:
## Live screen-text results (dev build, Screen Recording granted, toggle on)

| App | OCR blocks | text targets after AX explains | summary `Screen text:` |
|---|---|---|---|
| Google Chrome (Gmail) | many | 14 (glyphs, a URL, icons) | yes |
| Notes | — | 12 (sidebar labels, note body lines) | yes |
| Finder (list view) | — | 80 (file names, columns, dates) | yes, 2 KB |

What it fixed on the branch (`…` "never read this process's own windows"): the first run turned the Voice Control panel's own text — the user's instruction, our status line — into targets, because the capture reads every pixel in the window rect and the panel overlapped Chrome. Frames of windows owned by this pid now join the excluded frames.

Cost: the OCR pass runs inside the observe budget and pushed Finder to 3.1 s (partial). Follow-up before default-on: run OCR concurrently with the AX walk or give it its own budget; drop single-glyph blocks (`f`, `#`, `•`) before they become targets.

Denied-permission path not exercised live (access was already granted for the dev bundle).

